PR #12, opened 2026-04-18
Rewrite keyvalue item header with byte-aligned field layout

brayniac:
## Summary

- Adopt crucible's BasicHeader design: `[klen:1][flags:1][vlen:4]` = 6 bytes (12 with magic and checksum)
- Magic and checksum are enabled with the `integrity` feature which is on by default
- Replaces the old packed `[vlen|klen:4][flags:1]` layout that required cross-word bit manipulation
- Each field is directly byte-addressable — `klen` is a single byte read, `vlen` is a direct u32
- Rewrites `value.rs` (Value/OwnedValue) and `item/header.rs` with clean code
- Updates `item/raw.rs` to use `is_numeric()` instead of the old `value_type()` enum
- Header size changes: 5→6 bytes (without integrity), 9→12 bytes (with magic)


crates/keyvalue/src/item/header.rs
```diff
@@ -1,252 +1,155 @@
...
+//! │ KLEN │ FLAGS │             VLEN             │
+//! │ u8   │ u8    │             u32              │
+//! │ 8bit │ 8bit  │            32 bit            │
+//! └──────┴───────┴──────────────────────────────┘
+//!
+//! FLAGS: [is_numeric:1][reserved:1][olen:6]
+//!
+//! With `magic` feature, a u32 magic field (0xDECAFBAD) is appended:
+//!
+//! ┌──────┬───────┬──────────────────────────────┬──────────────┐
```
thinkingfish: I think the magic field should be at the beginning? It works as a OOB access guard.

brayniac: Done and added a checksum per discussion

PR #24, opened 2026-06-11
Fix CAS false-success via per-segment generation in a 64-bit token

brayniac:
## Summary

The current CAS token is the item's packed location truncated to u32. This has a false-success (ABA) case:

- Truncating the 44-bit location to u32 aliases locations whose segment ids differ by a multiple of 4096.
- Even at full width, a recycled segment can place the same key at the same offset, so a stale token matches a value the client never observed and the CAS falsely succeeds.

This follows the discussion on #23: rather than reintroducing a side counter (which becomes a two-word protocol once operations are concurrent), keep the location-derived token and fix its defects the way crucible's `CasToken` does.

## Changes

- Add a 16-bit `generation` counter to `SegmentHeader` (fits in existing padding; header stays 64 bytes), bumped in `reset()` — called only from `push_free`, which all recycle paths (eviction, merge, S3-FIFO, expiration, clear) funnel through.
- Port crucible's `CasToken`: a 60-bit token in u64 — generation in bits 59..44, location in bits 43..0. The token remains the content of the single hashtable slot word plus quasi-static segment metadata, preserving the single-word atomic swap path for future concurrent CAS.
- Widen `Item::cas()` and `Segcache::cas()` from u32 to u64, matching the memcache protocol's 64-bit "cas unique".

## Testing

- New `cas_stale_token_rejected_after_segment_recycle` test deterministically reproduces the ABA scenario (clear() recycles the segment LIFO; reinserting the same key lands at the identical 44-bit location) and asserts the stale token now fails with `Exists`. Precondition asserts verify the same-location setup, so the test fails loudly rather than silently not exercising ABA if free-queue ordering changes.
- Verified the test bites: with the generation bump removed, stale and fresh tokens are bit-identical.
- Ported crucible's CasToken layout tests; added a header generation unit test.
- `cargo test --workspace`, `--features debug`, loom suite (8/8 models), `clippy --all-targets --all-features -- -D warnings`, and `fmt --check` are all green; fuzz target updated and compiles.


thinkingfish:
Neat design. LGTM

thinkingfish (approved):

PR #37, opened 2026-07-23
Independent reimplementation, documentation, and repository hygiene

brayniac:
## Summary

Independently reimplements the core engine logic, refreshes documentation and tests, and moves the repository to a clean project-level licensing layout.

### Core engine
- Segment scan/merge/prune and eviction-segment ranking reimplemented from behavioral specifications, preserving the `&self` lock-free concurrency model. The memory-ordering-critical merge paths were validated by an adversarial concurrency review.
- Crate documentation, error messages, and functional test suites independently authored; the concurrency/roadmap tests are unchanged.
- Published algorithms cited where implemented: Segcache (Yang et al., NSDI 2021), S3-FIFO (Yang et al., SOSP 2023).

### Repository hygiene
- Remove the unused `datatier` crate (segcache allocates its heap directly via `memmap2`).
- Project-level licensing: `LICENSE-APACHE` + `LICENSE-MIT` + `NOTICE`, no per-file headers, holder "The cache-rs Authors".
- Distinctive integrity sentinels; version bumps (segcache 0.4.0, keyvalue 0.4.0, cuckoo-cache 0.3.0).

### Verification
`cargo test --workspace` (+`--features debug`) & doctests, `clippy --all-targets --all-features -- -D warnings`, `fmt` — all clean; loom model tests and the concurrent stress suite green.


crates/cuckoo/src/cuckoo.rs
```diff
@@ -359,18 +359,18 @@ impl CuckooCache {
     // Public API
     // -----------------------------------------------------------------------
 
-    /// Look up an item by key.
+    /// Fetch the item stored under `key`, if one is present and unexpired.
```
brayniac: Retrieve the item. Returns None if the item is not present or has expired.

crates/cuckoo/src/cuckoo.rs
```diff
@@ -467,18 +468,19 @@ impl CuckooCache {
         Ok(())
     }
 
-    /// Remove the item with the given key.
+    /// Delete the entry for `key`. The return value indicates whether a
+    /// live item was removed.
```
brayniac: This implies that we return false if the item is present but expired? That could use confirmation and clarification?

crates/cuckoo/src/item.rs
```diff
@@ -29,14 +29,14 @@ impl Item {
         self.raw.expire()
     }
 
-    /// Perform a wrapping addition on the value.
+    /// Add to the stored value, wrapping on overflow.
```
brayniac: We don't document failure cases for this or wrapping_sub ?

crates/keyvalue/src/item/header.rs
```diff
@@ -80,10 +86,12 @@ impl ItemHeader {
         }
     }
 
-    /// Check that the magic bytes match the expected value.
+    /// Assert that this header's magic bytes still match the sentinel
+    /// value written by `init`.
     ///
     /// # Panics
-    /// Panics if the magic bytes are incorrect, indicating data corruption.
+    /// A mismatch here means the header's memory has been corrupted, so
```
brayniac: "memory" might not be necessary here.

crates/keyvalue/src/item/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Items are the base unit of data stored within a cache.
+//! The packed on-heap representation of a single stored entry.
```
brayniac: Is it always on-heap? Consider rephrasing in-light of a desire to do tiered caching in the future.

crates/keyvalue/src/item/raw.rs
```diff
@@ -1,4 +1,4 @@
-//! A raw byte-level representation of an item.
+//! Field accessors over an item stored as packed bytes in segment memory.
```
brayniac: Another instance where we refer to "memory" unnecessarily?

crates/segcache/src/eviction/mod.rs
```diff
@@ -1,75 +1,116 @@
...
+    /// Reusable id buffer for `rerank()`. Kept around across calls so a
+    /// rerank never has to hit the allocator on the steady-state path —
+    /// only `clear()` + `extend()` into already-reserved capacity.
+    scratch: Vec<NonZeroU32>,
 }
 
 impl Eviction {
-    /// Creates a new `Eviction` which will handle up to `nseg` segments
-    /// using the specified eviction policy.
+    /// Constructs a new `Eviction` sized to manage up to `nseg` segments
```
brayniac: This wording is slightly awkward.

crates/cuckoo/src/cuckoo.rs
brayniac: Reworded to lead with the retrieval and state the `None` cases. (c306378)

crates/cuckoo/src/cuckoo.rs
brayniac: Confirmed against the impl: a present-but-expired item returns `false` (the expired slot is dropped, not counted as a removal). The doc now states this explicitly. (c306378)

crates/cuckoo/src/item.rs
brayniac: Both `wrapping_add` and `saturating_sub` now document the `NotNumeric` failure. (c306378)

crates/keyvalue/src/item/header.rs
brayniac: Dropped the unnecessary "memory". (c306378)

crates/keyvalue/src/item/mod.rs
brayniac: Good call re: tiering — dropped "on-heap" so the shared item type stays storage-agnostic. (c306378)

crates/keyvalue/src/item/raw.rs
brayniac: Dropped "segment memory" (now "a packed byte buffer"); also fixed a second unnecessary "memory" in the RawItem safety doc. (c306378)

crates/segcache/src/eviction/mod.rs
brayniac: Reworded to: Create an `Eviction` for up to `nseg` segments under `policy`. (c306378)

PR #45, opened 2026-08-18
feat: ziplist crate — compact collection block codec (list/hash/set/zset)

thinkingfish:
Pelikan's Redis-protocol server parses some thirty commands but can execute two, because the cache engine underneath stores exactly one flat value per key — collections have nowhere to live. This change gives them a home without abandoning the engine's item model: one compact block format that is the *permanent* representation of a collection, not a small-size optimization to be escaped later. The four types differ only in how entries are ordered and paired inside an identical block, so a single codec serves all of them, and the same block becomes the chunk of a chained (quicklist-style) collection later without a format break. Once this lands, cache-rs carries a frozen, verified byte format with typed read/write views — and nothing consumes it yet; the engine extension and the command wiring are the next two plans.

**What this asks of you: ratify a byte format that stored data will outlive. The freeze is the review.**

## What to look out for

Four ratifications, none blocking; carry these while you read — each is marked again where its evidence appears.

1. [ ] **Is the freeze surface complete?** Golden tests pin every uint tier and multi-byte varints. Decide whether anything else must be byte-pinned before data is stored durably.
2. [ ] **Integer-only sorted-set scores.** A deliberate Redis deviation (u64, no f64). The spec chose it; this PR turns it into bytes.
3. [ ] **Decode accepts non-minimal encodings** (a value writable in 1 byte arriving in a 3-byte tier decodes fine). Documented leniency to acknowledge, or reject now while it's cheap.
4. [ ] **The bounded proof.** One Kani harness runs tighter than sketched (immediate-tag entries, 40-byte buffer) with the excluded breadth covered by fuzzing/proptests instead.

## The mental model

This assumes nothing beyond "segcache stores key → bytes in append-only segments"; if you know the crate layout, skip to Where to look first.

A **block** is a self-contained byte encoding of one whole collection, sized to live inside a single cache item's value. Inside it, every element is an **entry** in one uniform codec — small integers cost 2 bytes, strings a tag plus length — and each entry carries a reverse-readable length so the block walks in both directions. A **typed view** (list/hash/set/zset) is not a different format: it is a convention about ordering and pairing over those same entries, enforced by the ops. Two contracts hold everything together: every mutation is **all-or-nothing** (it either fits in the caller's buffer or reports the exact size needed and touches nothing), and every traversal requires the buffer sliced to the block's **used length**, never its capacity.

```mermaid
flowchart LR
    A["RESP entrystore<br/>(pelikan — future Plan C)"] --> B["typed views<br/>list · hash · set · zset"]
    B --> C["splice engine<br/>all-or-nothing"]
    C --> D["entry codec<br/>tags + backlen"]
    D --> E["caller's buffer<br/>(a segcache item value)"]
```

Byte-level tables and a worked example live in `docs/ziplist.md` (checked in, byte-identical to the golden tests).

## What changed, and the claim it makes

One new crate (`crates/data-structure/ziplist` — nested for future siblings), its fuzz workspace, two CI jobs, and format docs; no existing crate is touched. The claim: any op sequence through the typed views leaves a block that re-parses cleanly and matches a reference model's state; refused writes leave the buffer byte-for-byte untouched; and the encoded form of every committed fixture can never change without a golden test failing. If the change is wrong, one of those three is false.

Citations below are pinned to `1fe821b` (rebased onto main after #47; file contents at cited lines unchanged).

## Where to look first

1. `crates/data-structure/ziplist/src/zset.rs:284-292` — the rev-range window mapping (`npairs-1-stop .. npairs-1-start`). This exact spot held the branch's one real bug (ZREVRANGE partial windows), found in review and fixed in `b48d025`; the regression test uses the Redis docs' own example.
2. `crates/data-structure/ziplist/src/block.rs:194-223` — `splice`, the single funnel under every mutation: the capacity bail-out **[3-adjacent: this is also where a non-minimal entry's size is simply whatever its bytes say]** and the `tail_off`/`nentry` arithmetic. Everything above it inherits its correctness.
3. `crates/data-structure/ziplist/src/hash.rs:218-245` — `hset` and the two-entry all-or-nothing insert; the torn-odd-pair state is the one this must make unreachable.
4. `crates/data-structure/ziplist/src/entry.rs:197` — `decode_backlen`, the backward varint walk over stored (i.e., untrusted-after-restart) bytes.
5. `crates/data-structure/ziplist/src/cursor.rs:237` — the test pinning the used-length contract: a stale-but-valid entry planted past the tail is invisible through a correctly sliced buffer. The contract is documentation-enforced, not type-enforced — judge whether that's enough until Plan C.

Safe to skim: `tests/golden.rs` and `tests/model.rs` bodies (mechanical once you've read one), `benches/`, the fuzz targets' `Arbitrary` plumbing, and `docs/ziplist.md`'s tables **[1: but decide whether the golden set — every uint tier at boundaries, a ≥128-byte string exercising 2-byte length-varint and 2-byte backlen, one fixture per type — is the freeze surface you want]**.

## Testing

- Unit + golden + model: `cargo test -p ziplist` → **49 pass, 0 fail** (35 unit, 6 golden, 8 model). Model tests run 512 cases/type against `VecDeque`/`BTreeMap`/`BTreeSet` references with full-state equality after every op; small-buffer variants assert refusals actually fire (a counter, not hope).
- Fuzz: 3 targets — `decode` (arbitrary bytes), `ops` (splice engine), `typed_ops` (per-type APIs, fuzzer-controlled keys/values/deltas, varied 64–4096-byte buffers). `cargo fuzz run <t> -- -runs=100000` each: **zero findings**. CI runs short smoke iterations only.
- Kani: `cargo kani -p ziplist` → **4/4 harnesses verified**. Harness 4 is bounded tighter than the plan sketched **[4: bounds and the CBMC memory-exhaustion history that forced them are documented in `src/kani.rs`; strings/multi-byte tags/longer blocks are covered by fuzz+proptest instead of proof]**.
- Benches: `cargo bench -p ziplist -- --test` compiles and runs; **no baseline numbers recorded** — they exist to price Plan B's in-place-vs-rewrite ablation later.

Gaps, stated: zset range/count reads are unit-tested and fuzzed but not property-tested against a model; there is no differential test against real `redis-server` (deferred to Plan C where a server exists to speak to); decode's acceptance of non-minimal encodings is documented but has no test asserting it **[3: the committed behavior is lenient; a strictness test would be the reversal]**; scores' integer-only deviation **[2]** is enforced at the API type (`u64`) so no test can exercise a float path.

## Judgment calls

- **Integer u64 scores** **[2]** — rejected f64 (NaN/±inf semantics, lossy client round-trips, and cache workloads score by timestamp/ID). Evidence that would change it: a real workload needing fractional ranks. Reversal is a new zset `format` byte, not a break.
- **Sorted bodies, no intra-block index** — rejected per-entry offset tables (~2B/entry) because the chained design's directory provides the coarse index and block size bounds scans. A measured lookup bottleneck on realistic blocks would reopen this.
- **Pops take a closure** (`pop_front(|v| ...) -> Option<R>`) — returning the value is unimplementable without alloc: it borrows bytes the removal memmoves (compiler-verified, E0502). An `alloc` feature would change it.
- **`hincrby`/`zincrby` auto-vivify at 0+delta** — matches Redis; rejected erroring on missing field.
- **Non-minimal decode leniency** **[3]** — rejected strict canonical validation on every decode (cost on the read path; this crate is the only writer). Becomes wrong the day encoded-byte identity matters (checksums, CAS-on-bytes).
- **Hash values stored verbatim as strings** even when numeric, while `hincrby` writes canonical integers — both forms read identically; frozen by goldens; costs ~2× on numeric hash values. Revisiting is a hash `format` bump.

## Production-only risks

- **The format freeze itself.** Blocks stored through file-backed datatier persist across binary upgrades; an encode/decode asymmetry ships silently today and detonates on restart with old data. Surfaced by: the golden tests (tripwire at CI time) and, once Plan C lands, a `collection_decode_error` counter at runtime. **This is the risk this PR asks you to accept** — the compensating control is the freeze discipline, so weigh item [1] with it.
- **Kani CI memory headroom.** Harness 4 neared 30GB during development at wider bounds; a future edit to splice/decode internals could push the CI job into OOM. Surfaced by: the `kani.yml` job failing with CBMC's memory diagnostic (first suspect documented in `src/kani.rs`).
- **CI fuzz depth.** Smoke runs are minutes; the 100k-run clean result is a local, one-time fact. A scheduled deeper fuzz job is the observable if we want one.

## Not in scope

Engine-side in-place update with capacity slack (Plan B) and the pelikan RESP entrystore (Plan C) — including typed wrap-existing-buffer constructors, which deliberately do not exist yet; the carry-list lives in the pelikan spec (`docs/superpowers/specs/2026-08-18-ziplist-collections-design.md`, branch `spec/ziplist-collections`).


brayniac:
> **Automated review.** This was produced by Claude Code running a multi-agent review pass over the diff, then re-verified against the branch by hand before posting — the claims below cite the lines they rest on, so please check them rather than trusting them. Two of the seven review angles (conventions, and a cross-check/build pass) were lost to a rate limit partway through, so coverage is uneven; CI covers the build side of that gap. Judgment calls on the four ratification points are stated as recommendations, not verdicts — they're yours to make.

## Review: ratify with five small pre-merge changes

Reviewed at `4b9197b`. The codec itself is in good shape — I hand-checked the tier boundaries (250/251, 2^16, 2^24, 2^56), the forward-varint and backlen boundaries (127/128, 16383/16384), and `splice`'s `tail_off`/`nentry` maintenance across insert/remove/replace × head/mid/tail, and found no arithmetic disagreement between encoder and decoder. The backlen recompute-and-compare tripwire, the all-or-nothing `NeedBytes` contract, and the version story (per-type format byte + must-be-zero reserved flags + error-on-unknown) are all sound. What follows is almost entirely about the *freeze surface* rather than the code.

Taking the PR at its word that the freeze is the review, I've sorted findings by reversibility: anything that stored bytes will outlive comes first, everything else is post-merge work.

### Must decide before merge (frozen by stored bytes)

**1. The tier goldens can't detect the flip they were written to catch.** `tests/golden.rs:174-206` says in its own doc comment that it exists so "an endianness flip or a tier-boundary regression in `tag_and_data_len` would pass every other committed test." But all four vectors use all-ones payloads — 65535 → `FB FF FF 03`, 2^24−1, 2^56−1, `u64::MAX`. Those are byte-palindromes: changing `to_le_bytes` to `to_be_bytes` at `src/entry.rs:241` passes every committed test in the crate (the unit tests at `entry.rs:387-403` assert sizes only; everything else round-trips, and round-trips are blind to a symmetric change on both sides). One asymmetric value per tier closes it — e.g. `0x1234` → `FB 34 12 03`, `0xABCDEF` for u24, and distinct-byte values for the 56- and 64-bit tiers. This is the single most valuable ten minutes in the PR.

**2. `FLAG_CHAIN_ROOT` is accepted by v1 but means nothing to it.** `header.rs:7-8` sets `KNOWN_FLAGS = FLAG_CHAIN_ROOT`, so `parse` accepts bit 0 — but nothing in `src/`, `tests/`, `fuzz/`, `benches/`, or `examples/` ever sets or reads it; every block is written with `flags = 0`. The reserved-flag rejection at `header.rs:67-70` exists precisely so a v1 reader refuses data it can't interpret, and bit 0 has been punched out of that protection in advance for a feature that doesn't exist yet. The consequence lands when chaining ships: a v1 reader hands back a chain root as if it were a complete standalone collection — a silently truncated collection instead of an error. Suggest `KNOWN_FLAGS = 0` here, and introducing the constant in the PR that implements chaining. It is free today and unavailable afterward.

**3. Non-minimal encodings: reject now.** I'd ratify this one in the strict direction, for a cost-asymmetry reason rather than a purity one. This crate is currently the only writer of these bytes and its encoder is always minimal, so **no legitimately stored block is non-minimal today — tightening now invalidates nothing and costs ~2 comparisons on a path (`validate`'s entry walk, `block.rs:33-67`) that already touches every byte.** Tightening after merge is a behavior change against whatever a second writer or corruption has by then stored. The format is also already half-strict in a way that's hard to defend as intentional: the backlen *is* validated by recompute-and-compare (`entry.rs:297-309`), while the tier arm (`entry.rs:276-289`) and `decode_varint_len` (`:137-157`) accept any width. Worth noting the PR description's "documented leniency" is a touch generous — the behavior appears only in the journal's deferred list (`docs/journal/2026-08-18-ziplist-codec.md:92`), not in `docs/ziplist.md` and not in any test, so today it's frozen by accident rather than by decision. If you'd rather keep the leniency, that's defensible, but it needs to be normative text plus an acceptance test; the current silent-and-untested state is the worst of the three options.

**4. Two format elements are pinned only by round-trips.** The empty-block encoding (`src/header.rs:119-128` does `init_empty` → `parse` and checks fields) will be the most frequently stored shape in the system once collections exist, and a paired change to `write_to` + `parse` would pass. Same for the ≥3-byte forward varint (`entry.rs:406-411` round-trips a 20,000-byte string with no byte assertion). Both are one golden vector each.

**5. There are two Redis deviations in the zset bytes, and only one is on the ballot.** Beyond u64 scores, equal-score tiebreak uses `compare` (`entry.rs:329-336`), so canonical-numeric members sort numerically and ahead of all strings, rather than Redis's pure byte-lex order. With tied scores Redis orders `"10" < "9"`; this format orders `9 < 10`, and `"10"` before `"a"`. That's stored order, so it freezes with everything else, and `ZRANGE` output will visibly differ for numeric members. It should be ratified explicitly alongside the score decision rather than discovered later.

On the score decision itself: **ratify u64.** The escape hatch is real and well-built — `parse` rejects any `format != 0` (`header.rs:63-65`), so a future `Zset/0x01` with i64 or f64 scores is a clean extension in which old data stays readable and unknown data errors rather than misparses. One fact worth stating in the spec rather than leaving to be discovered: the entry tag space is fully saturated (0–255 all assigned, `entry.rs:34-38`), so any future score encoding *must* arrive as a format bump; there's no spare tag. I'd also record that negatives, not fractions, are the likelier real-world ask (the "recency = −timestamp" idiom), and that `zincrby` below zero returns `IncrError` (`zset.rs:430-460`) as the protocol-visible consequence.

### One correctness gap, and it's a shape this repo just paid for

**6. The format is u32-bounded, and nothing enforces it at the buffer boundary.** `encode_backlen`'s own doc states the reasoning: "Bound is u32::MAX: a tag_plus_data span is u32-bounded by the header format (`tail_off: u32`)" (`entry.rs:165-167`) — but that bound is a `debug_assert`, and `splice` writes `self.hdr.tail_off = new_tail_off as u32` (`block.rs:271`) after a fit check that bounds `new_used` against `self.buf.len()`, not `u32::MAX`. Nothing rejects an oversized buffer at construction or parse. On a >4 GiB buffer, `tail_off` wraps silently: every individual op returns `Fit`, and a later re-parse reports `Corrupt` on a block the writer was told it had written successfully. `cursor.rs:93` has the matching truncating comparison (`self.off as u32 > hdr.tail_off`), which defeats its own misplaced-cursor guard above 4 GiB.

Not reachable in the intended deployment — segcache items cap well below this — so I'd call it latent rather than live. I'm flagging it anyway because it is exactly the defect we just fixed in segcache #79/#86: a field-width ceiling that the docs assert, the code relies on, and nothing enforces, failing by silent truncation rather than by error. The remedy is the same one-liner: refuse a buffer larger than `u32::MAX` where `BlockMut` is constructed, derived from the header field rather than a literal, so the ceiling is structural instead of a precondition each caller has to honor.

### Worth doing, but nothing here should hold the merge

These are all internal API and can change any time, so I'd take them as follow-ups rather than growing a 6,700-line PR:

- **The classification rule exists four times.** `canonical_uint(bytes).map_or(Str, Uint)` is a private fn in `hash.rs:42`, `set.rs:28`, and `zset.rs:56`, plus inlined in `compare_raw` (`entry.rs:343-344`). It's the crate's most load-bearing convention — it decides both storage classification and sort order — so a future tweak applied to three of four sites makes lookups disagree with stored order, i.e. silently lost members. One `classify()` in `entry.rs`, with `compare_raw` defined through it.
- **Related, and worth a look before it hardens:** `canonical_uint` (`entry.rs:43`) is a semantic twin of the workspace's existing `keyvalue::numeric::parse_simple_numeric`. ziplist is `no_std` and keyvalue isn't, but `numeric.rs` is core-only, so this could be shared rather than forked — otherwise the workspace ends up with two definitions of "simple numeric" that can drift apart.
- **Pair operations are two splices where one would do.** `insert_pair` (`hash.rs:338`, duplicated near-verbatim in `zset.rs:483`) memmoves the tail twice and rewrites the header twice; `hdel`/`zrem`/`rescore` do the same on removal. Since the two entries are always adjacent, a `BlockMut` pair primitive would halve the memmove traffic on every HSET-new/HDEL/ZADD *and* dissolve the "remove the higher-offset cursor first" rule that's currently a comment repeated at three call sites. That last part is the house-rule argument: it turns a discipline into a structure.
- **Every visited entry is decoded twice** — `Cursor::at`/`next` call `decode()` and discard the value, then `cur.value()` decodes again (`cursor.rs:69,101`), so an O(n) scan pays 2n full decodes including the backlen tripwire each time. Returning the decoded value alongside the cursor would halve it.
- **`varint_len_for_backlen` (`entry.rs:222`) is byte-identical to `varint_len` (`entry.rs:104`)** — and `decode`'s corruption tripwire depends on the two agreeing, so this is a duplicate that must not drift.
- **`locate` returns `Corrupt` for a merely out-of-range index** (`cursor.rs:118`), conflating a bad argument with bad data. Any storage layer whose policy is "Corrupt ⇒ quarantine the key" would discard a healthy collection because a caller asked for index `nentry`. There's no `OutOfRange`/`NotFound` variant, so it's structural rather than a wrong pick.
- Smaller: `PairKey` is a single-variant enum whose parameter is discarded (`map.rs:39,90`); `zscore`/`sismember` return `Result` that can never be `Err`; `Cursor`'s pub `off`/`len` let callers fabricate cursors while staleness is doc-only.

### On the Kani harness

Adequate for a freeze, with one doc fix. The harness is honest about itself — `src/kani.rs:139-148` explicitly says the tail-only check is narrower than a full re-validation — and the excluded breadth is genuinely covered elsewhere: `fuzz_targets/ops.rs:91-148` exercises arbitrary `u64` uints and strings through insert/remove/replace at arbitrary positions with a full `Block::parse` after **every** op, which is exactly the multi-entry re-walk the proof drops. Harnesses 1–3 already prove the entry codec over all `u64`, so tier arithmetic is covered, just not through splice.

The one mismatch: `docs/ziplist.md:293-296` describes harness 4 as "buffer <=96 bytes, entries restricted to `Uint`", but the shipped harness uses a 40-byte buffer and immediate tags only (`v <= 250`) — materially narrower than "Uint". Two-line fix. Also worth correcting in the safe direction: CI runs `-runs=100000` per fuzz target (`.github/workflows/fuzz.yml:19,61`), not the "short smoke iterations" the PR describes. Since the fuzzing is the load-bearing compensating control for everything the proof excludes, that depth is worth stating as deliberate.

### One ambiguity the freeze doesn't resolve

Generic `Block::parse` accepts a Hash block with odd `nentry`, or unsorted or duplicate fields — `validate` (`block.rs:33-67`) checks entry-chain, `tail_off`, and `nentry` consistency, and nothing else. So the format currently doesn't say whether such a block is *corrupt* or merely *unconventional*. That question lands squarely on the wrap-existing-buffer re-validation decision that `docs/ziplist.md:232-246` defers to Plan C, and it's the kind of thing two implementations will answer differently. One normative sentence now — either "body conventions are mandatory; violating blocks are corrupt" or "advisory; readers must tolerate" — costs nothing and prevents that.

### Summary

Ratify, with five small changes first: asymmetric tier goldens and an empty-block golden (1, 4); `KNOWN_FLAGS = 0` (2); strict non-minimal rejection, or normative documentation plus a test if you prefer leniency (3); the `u32` buffer guard (6); and spec text covering both zset deviations plus the corrected harness bounds (5, Kani). Everything in the third section is follow-up work.

Nice things that need no change and are easy to miss: the per-type format byte plus must-be-zero reserved flags gives a genuinely clean upgrade path; endianness is stated normatively in `docs/ziplist.md:31`; the backlen recompute-and-compare is a cheap and well-placed corruption tripwire; and `NeedBytes`'s all-or-nothing contract is respected everywhere I checked, including the failure paths.

---

https://claude.ai/code/session_01GWv9oJmijF6kkQQh6Ac8h7

PR #46, opened 2026-08-19
segcache: de-duplicate fresh-key inserts with striped hashtable insert locks (7f follow-up)

brayniac:
Closes the item-7f tracked follow-up: the concurrent fresh-key duplicate-publish race.

## What

- **Closes the fresh-key duplicate-publish race.** All hashtable entry *creation* (empty-slot claim, ghost takeover) now runs under a per-key-hash stripe (1024 `CachePadded<Mutex<()>>`, 16 under loom) with an under-lock absence re-check, so at most one live hashtable entry per key exists at all times. `insert` is the only creation path in the table, so the invariant is enforceable at one layer. The user-visible fix: a deleted key can no longer resurrect with a losing insert's value.
- **Fixes a latent single-threaded duplicate too.** `insert` now scans *all* candidate buckets for a live entry before any claim; previously a key whose entry lived in its second-choice bucket could be shadowed by a fresh claim in a freed first-choice slot — no race required (`test_replace_across_buckets_no_duplicate`).
- **Plus an insert-vs-relocation fix:** `try_replace_existing` re-validates the slot word when `verify` fails, so a stale `packed` from a concurrent same-key relocation is retried instead of being misread as "a different key occupies this slot" (which would have created a duplicate).
- **Overwrites, `get`, `delete`, `cas`, and numeric ops are untouched** — the lock-free replace fast path returns before the lock is taken. The stripe is a leaf lock (`LOCK: insert-stripe`): its critical section is bucket-word CAS plus verifier memory reads, never another lock, pin, or wait.

## Coverage

- New hashtable-layer race test (2000 trials x 4 threads, bite-checked red/green against the removed lock) and a public-API resurrection stress test (Test 10, 1000 trials).
- Two loom models (plain fresh-key race and ghost-takeover variant) — a mutex-serialized invariant is SC-independent, so loom genuinely verifies it, unlike the SeqCst Dekker pairs. Loom suite now 21/21; multi-key models assert stripe distinctness so a hash change can't silently shrink the explored interleaving space.

## Benchmarks

Same machine, identical bench code on both sides (the new `set_fresh` unique-key miss-fill bench landed as the first commit, before the fix):

| bench | pre | now | change |
|---|---|---|---|
| `set_fresh/8b/64b` | 163.9 ns | 143.1 ns | **-11.7%** |
| `set/1b/64b` | 58.3 ns | 56.5 ns | -2.3% |
| `incr/hot_counter` | 55.7 ns | 53.5 ns | -4.1% |
| `get/1b/0b` | 17.0 ns | 17.1 ns | within noise |

The fresh path gets faster despite the lock because the old claim path's redundant `min_by_key`/`count_occupied` probe and `Vec` allocation are gone.

## Known, unchanged, and now more frequently hit

The drain-vs-`raced_old` decrement gap (an entry unlinked without a remover pin when a fresh insert resolves to a replace) is unchanged and deliberately out of scope per the design spec §4 — but note its *hit rate goes up*: post-fix, the loser of every fresh-key race funnels into the `Ok(Some(raced_old))` arm. It remains accounting-benign (transient over-count, self-healing on segment recycle; the drain owns the segment's accounting wholesale) and is documented in `segcache.rs` and `segment.rs`.

Design spec and implementation plan are included under `docs/superpowers/`.

## Adversarial review

Three independent adversarial reviewers examined the branch's risk areas before merge. Results:

**Fixed in this PR** (final commit):
- **Same-slot CAS retry in `try_unlink_in_bucket` / `try_to_ghost_in_bucket` / `try_cas_in_bucket`.** A racing reader's freq-bump CAS could spuriously fail these primitives, which then abandoned the entry — letting `Segment::clear` recycle a segment with a still-published entry (breaking the "unlinked before recycle" invariant that both the reader re-lookup and this PR's verify-fail re-validation rely on), making `delete` spuriously return `false`, and aborting merge relocations. Pre-existing (same class F4 fixed for the insert path; three primitives were left behind), and **live, not latent**: the new native stress test failed within ~1 trial pre-fix. Now mirrors the F4/`cas_location_at` retry (bounded: freq saturates at 127), pinned by a new loom model (`loom_remove_vs_freq_bump_unlinks`, verified red pre-fix) and a native tripwire. Loom suite: 22/22.
- **Insert-stripe poison recovery**: `lock().unwrap_or_else(|e| e.into_inner())` — the stripe guards `()` and every mutation in its critical section is a single CAS, so a panicking inserter can't half-write state; poisoning must not permanently kill 1/1024 of the keyspace. (The `chain_lock`/eviction mutexes keep `.unwrap()` deliberately — they guard real multi-field state.)

**Verified safe** (reason recorded):
- *Creation exclusivity*: the only `0→live`/`ghost→live` transitions in the crate are `try_claim_new_slot`'s three passes, all under the stripe; every other slot write is in-place mutation or destruction. Ghost/eviction/delete races against the under-lock section all resolve to single-entry.
- *Memory orderings*: every publish is a Release CAS; every load followed by a segment-byte read is Acquire; the mutex gives the HB edge for under-lock re-checks; the re-check's obligation (observe *some* live entry if one exists) survives in-place relocation because entries never migrate slots.
- *Retry termination*: all same-slot retry loops are bounded — every retry is paid for by another thread's successful CAS, and freq bumps saturate (probabilistic >16, hard cap 127).
- *Error paths*: the stripe guard is dropped before any `rollback_reservation`/`remove_at`/`chain_lock`; the critical section is a genuine leaf (bucket-word CAS + raw verifier reads).
- *Regression sweep*: bench file frozen at commit 1 (A/B trustworthy), metrics call sites byte-identical, all feature combos (`debug`, `loom`, `--no-default-features`) compile, no SIMD/prefetch code touched, stripe array is ~128KiB/instance with no bulk-construction site affected.

**Pre-existing findings, tracked as follow-ups** (not introduced here; filed so they don't ride silently):
- #49 — `insert` can deadlock spinning on `try_pin_remover` while holding its reservation's `WriterPin` (drain waits on `active_writers` before unlinking → cycle). This PR widens the pin-hold window slightly (the stripe is a blocking point) but adds no cycle.
- #50 — the raced-old/rollback unpinned unlink can pin a *recycled-and-reused* segment incarnation (`Location` carries no generation) → misdirected decrement or `max_item_offset` assert. This PR corrects the in-code comment to stop claiming the pin can only fail.


thinkingfish:
## Review

The core mechanism holds under verification: `insert` is the only entry-creation path, the stripe-lock claim is sound (`SegmentsVerifier` performs only buffer reads), every removed behavior is re-established, and return-value contracts are preserved. Nine findings, ranked by severity. Five further candidates were examined and refuted; they are listed at the end so the ground covered is visible.

### Correctness

1. **`segcache.rs:318` — the `Ok(Some(raced_old))` arm can decrement an item in a reused segment.** The arm pins after unlinking, and `reserved` is dropped at line 315, so a drain can seal, drain, and recycle the segment before the pin attempt. `Location` carries no generation bits and `try_pin_remover` checks only `state_is_removable`, so the pin can succeed on the wrong incarnation, and `remove_item_at` then parses garbage at the stale offset: a large bogus size trips the `live_bytes()` assert (a plain `assert!`, which panics in release builds); a small one corrupts accounting; if `live_items()` reaches zero, the empty-free path clears a live segment. The rewritten comment states that the pin fails and the drain owns the accounting wholesale — that claim is false, and this is the one place the PR makes the surface worse rather than better. At minimum restore the earlier comment wording; the real fix is to re-validate the location before `remove_at` on this path, the way the replace arm pins before unlinking (lines 258-268).

2. **`table.rs:858` — `try_unlink_in_bucket` (and `try_to_ghost_in_bucket`, line 893) advance past the slot on a matching-slot CAS failure.** This is the bug class the PR fixes in `try_replace_existing`, surviving in the delete family. A concurrent `get` bumps the slot's freq bits on every hit below saturation; the delete's CAS fails on those bits alone; `continue` skips the still-matching slot; `delete` returns false while the key stays live. A client observes DELETE → not-found, then GET → hit. Pre-existing at the base commit, but the PR rewrote this function family and established same-slot retry in the sibling passes — apply the same pattern here.

3. **`table.rs:824` — the any-ghost claim pass skips the slot on CAS failure**, unlike the matching-ghost pass at line 787. A concurrent ghost-freq bump can turn an insert with a claimable ghost into `Err` → `rollback_reservation` → `SegcacheError::HashTableInsertEx`, with no retry at any layer; with `num_choices == 1` the sorted retry is skipped entirely (line 1198), so one lost CAS suffices. Wrap this pass in the same re-read-this-slot loop as the first pass.

### Tests and benchmarks

4. **`table.rs:1713` (also 1974-1991) — the loom stripe-distinctness asserts depend on ahash stability that ahash disclaims.** The table seeds are fixed, but ahash states its output differs across versions and CPU-feature builds: a routine `cargo update` reshuffles every stripe (the three-key test trips one of its pairwise asserts in roughly 18% of reshuffles), and `-C target-cpu=native` on x86_64 changes the stripes today. Select the keys by brute force at test start, as `test_replace_across_buckets_no_duplicate` already does for buckets, through one shared helper.

5. **`benchmark.rs:121` — measured: 2.6% of `set_fresh` iterations take the error path at steady state**, swallowed by `let _ =` at line 139. The comment's premise — that hash_power 20 exceeds what the 64MB heap holds live — is false: about 762K live items against 2^20 slots is 72% table load with 8-slot buckets. The headline −11.7% therefore blends the claim path with the reserve → define → `Err` → `rollback_reservation` path. `hash_power(21)` measures at 0.005%. Raise the power, count the errors, and assert the rate is near zero so contamination cannot recur silently.

### Cleanups

6. **`table.rs:1156` — drop the lock-free phase-A replace scan, or extract the scan both phases share.** `Segcache::insert` reaches `hashtable.insert` only on the `lookup_slot` miss arm, and `try_replace_existing` skips ghosts, so phase A can find an entry only in the raced-publish window the under-lock re-check (1180-1186) must catch anyway; every production fresh insert pays a redundant candidate-bucket scan. The two byte-identical loops are held together only by "change both together" comments — an edit that lands in one loop and not the other weakens the absence check that prevents the duplicate-entry bug this PR fixes. One shared helper makes the identity structural.

7. **`table.rs:1196` — the sorted least-full retry is a vestige.** The first pass is exhaustively greedy per bucket, so the retry succeeds only when a concurrent remove frees a slot within microseconds — and then order is irrelevant across at most 8 buckets. The sort key (`count_occupied`) excludes ghosts, which are claimable, so the ordering is wrong besides. A plain second claim loop in choice order keeps the retry behavior and deletes the scratch array, the copy, and `count_occupied`'s last caller (line 1202).

8. **`table.rs:2126 and 2192 — the live-entry counting oracle exists in three copies**: `count_live_entries` (line 1331) plus two ~28-line clones in the new loom tests, already divergent (the clones dropped the verifier call). Hoist `count_live_entries` to file scope under `#[cfg(test)]` — its body uses only `ht.bucket`, `Hashbucket` methods, and `crate::sync::Ordering`, so it compiles under the loom cfg — and call it from both tests.

9. **`table.rs:765` — the stripe-lock precondition is enforced by prose.** A future caller can invoke `try_claim_new_slot` without the lock; the change compiles and silently reopens the duplicate-publish race, detectable only by probabilistic stress tests. Take a `&MutexGuard<'_, ()>` (or a `StripeGuard` newtype with a `debug_assert` on the stripe index) so the requirement is type-checked. About five lines.

### Examined and refuted

Stripe-mutex poisoning (no unwinding-panic path under the guard); ghost-stranding freq loss (dormant machinery, documentation accurate, identical at base); the stale-verify recycle invariant (holds on every byte-reuse path); an unbounded verify-retry spin (exit requires one stable re-read; sustained looping needs a same-tag slot rewritten inside every sub-microsecond window, and each retry reflects completed external progress); a conventions violation (none — the clocksource rule concerns internals, and `Segcache::insert`'s public signature already takes `std::time::Duration`). Below the severity cut: the 100k-iteration loop test setup and the 2000-trial stress sizing.

Issue #50, opened 2026-08-19
segcache: raced-old/rollback unpinned unlink can pin a recycled segment incarnation (Location has no generation)


brayniac:
Found by the pre-PR adversarial review for #46. Pre-existing (the raced-old arm shipped with 7f); #46 corrects the comment to stop claiming the pin can only fail.

**The hazard:** in `Segcache::insert`'s fresh-key arm, `hashtable.insert()` returning `Ok(Some(raced_old))` means the racer's entry was already unlinked *inside* the hashtable call, before any `try_pin_remover`. `try_pin_remover` gates on state alone (`Sealed | Live`) and `Location` carries no generation, so between the unlink and the pin the segment can complete `Draining -> clear -> recycle -> reserve -> Live`. The pin then succeeds on the **new incarnation**, and `remove_at(raced_seg, raced_offset)` either trips `assert!(offset <= max_item_offset())` (panic) or decrements the wrong incarnation's `live_items`/`live_bytes` (which feed the empty-free fast path and merge compact ratio). `rollback_reservation` has the same shape. The replace arm is immune (pin happens before its `cas_location_at`, which fails closed).

**Suggested direction:** eliminate the unpinned unlink instead of patching it — make the hashtable's fresh path return "exists" *without* unlinking when the under-lock re-check finds a racer's entry, and let `Segcache::insert`'s outer loop retry into the pin-first replace path. That also closes the long-tracked missed-decrement gap in the same arm. (A generation snapshot after `insert()` returns narrows but cannot close the window — the recycle can precede the snapshot.)

See the "Adversarial review" section of #46.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FuDgaMXJWz1fnYGhYF7TLh

brayniac:
New argument for this issue, with a measurement behind it.

Generation-tagged locations are not only ABA-narrowing — **they are what would convert ThreadSanitizer from a manual target into a viable CI gate.**

[...]

PR #54, opened 2026-08-19
segcache: writers never wait on drain-claimed segments (fixes #49)

### LINKED ISSUE #49: segcache: insert can deadlock spinning on try_pin_remover while holding its reservation's WriterPin
Found by the pre-PR adversarial review for #46. Pre-existing (shipped with the 7f concurrent-write work), not introduced by that PR.

**The cycle:** `Segcache::insert`'s replace arm spins `try_pin_remover(old_seg) else { backoff.snooze(); continue }` while still holding the reservation's `WriterPin` (deliberately — publish happens under the pin). A drain that has already won its `Sealed -> Draining` claim CAS waits `while active_writers() != 0` *before* unlinking anything, and `try_pin_remover` rejects `Draining`. Two reachable shapes:

1. **Same segment:** thread A holds `WriterPin(S)` for its new reservation while the key's previous copy also lives in S (hot key written twice into the same tail). S gets sealed (the chain seal does not check `active_writers`), an evictor claims S for drain and waits on A's writer pin; A spins forever on `try_pin_remover(S)` because S is `Draining` and its entry is never unlinked (the drain is blocked before `finalize_drained`).
2. **Cross segment:** A holds `WriterPin(S1)` spinning for S2; B holds `WriterPin(S2)` spinning for S1; drains claim both. Same cycle, two hops.

`replace_at` has the same shape. The `get_item_frequency(..).is_none() -> Exists` escape hatch does not help: the blocked drain hasn't unlinked yet, so the entry still resolves.

**Suggested direction:** never spin on a drained segment while holding a `WriterPin` — bound the retry and fall back to `rollback_reservation` + error, or release the reservation's pin before waiting.

Note: #46's insert-stripe mutex adds another blocking point inside the pin window (no new cycle — the stripe is a leaf — but the hold time grows). See the "Adversarial review" section of #46.

brayniac:
Fixes #49 — the writer-vs-drain deadlock found by #46's pre-PR adversarial review.

## The cycle

`Segcache::insert`'s replace arm and `replace_at` (the cas publish path) spun on `try_pin_remover(old_seg)` while holding their reservation's `WriterPin`. A drain that has won its `Sealed -> Draining` claim waits `while active_writers() != 0` *before* unlinking anything, and `try_pin_remover` rejects `Draining` — so a writer whose key's old copy lives in a claimed segment (same-segment: a hot key written twice into one tail; or crossed between two writers and two drains) wedges forever.

## The rule

**A writer holding a `WriterPin` never waits on a segment it cannot pin.** Pin failure falls through to an *unpinned* `cas_location_at` slot swap — the CAS validates the exact packed word, so a stale slot fails safe — and on success the `remove_at` decrement is skipped: whoever owns the segment settles its accounting (`delete()`'s pin-fail arm already worked this way). `claim_for_drain`'s termination argument is now global: no counted writer can be parked on the segment being claimed, including the same-segment case.

The state split (documented at both sites):
- **Draining** — a drain owns it; `clear` + recycle reset the header counters wholesale.
- **Relinking** — production-real via `cas` on a merge-destination republish (the cas-path variant of the cycle: the merge owner holds the destination in `Relinking` across its whole candidate loop while waiting on `active_writers`). No drain owns it *yet*: the skip persists as a bounded over-count until that destination's own eventual drain — conservative direction only (the segment merely looks fuller).
- **AwaitingRelease** — unreachable here: `clear` unlinks every entry before condemn (reliable since #46's same-slot-retry fix).

## Gauge settling (and a pre-existing s3fifo metrics bug)

Skipping `remove_item_at` also skipped the only decrement of the process-global `ITEM_CURRENT`/`ITEM_CURRENT_BYTES` gauges, which no recycle resets — permanent drift, made routine by the Relinking case. Rather than an unpinned read of the old item's size (a genuine byte-level data race — the review's first suggestion, withdrawn on the implementer's analysis), `Segment::clear` now settles the residual: the gauges move in exact lockstep with the header counters, so whatever the header still counts at clear time *is* that segment's drift, subtracted race-free under drain ownership. This heals every unpinned-unlink source (`raced_old`, rollback, `delete`'s arm), not just this PR's two. Verifying the lockstep premise exposed a **pre-existing, independent bug on main**: `s3fifo_promote_from` re-homed an item's header count without re-adding the gauges (unlike `copy_into`'s batch compensation) — a permanent under-count, now fixed.

Also: `Segment::check_integrity`'s exact item-count match relaxed to the directional invariant (hashtable-resolving count may not *exceed* the header counter; over-count is the documented accepted gap), preventing a latent `--features debug` CI flake; `try_reserve`'s now-tautological write_offset tripwire dropped in favor of the real guards.

## Tests

Two deterministic regression tests, both verified to WEDGE pre-fix (10s watchdog) and complete in 0.00s post-fix, independently bite-checked per site:
- **Test 11**: claim a sealed segment via the drain-claim test hook, never finalize (a blocked drain, distilled), overwrite a key living in it — pre-fix `insert` spins forever.
- **Test 12**: park a sealed segment in `Relinking` (the same transition `link_dest_at_head` performs; a drain-claimed segment isn't *readable*, so `cas` can't reach the wedge through it — the plan's original scaffold was impossible, which is itself evidence the Relinking path is the real cas-side exposure), then `cas` the key — pre-fix `replace_at` spins forever.

Gates: full workspace tests, debug + release, loom 22/22, both clippy invocations, fmt — all clean.

## Adversarial review

An adversarial pass over the finished branch (five risk areas):
- **Fixed:** the gauge drift (above); the s3fifo lockstep hole its verification surfaced; the latent `check_integrity` flake; stale comments (`cas_location_at`'s pin-based justification, an unreachable `release_unused` path claim, a dead "load-bearing retry" claim).
- **Verified safe:** exactly-zero-or-one decrement in every interleaving (every decrementer is gated on winning a location-checked unlink of the exact packed word — a skip can never combine into a double, `live_bytes >= 0` unreachable); no merge/s3fifo bookkeeping assumes its republished entry survives; the write tail can never be a merge destination; cas token semantics unchanged (the removed early `Exists` check is exactly subsumed by the post-failure check); no false-positive path for the touched debug assertions; Test 12's parked state is production-reachable with identical surrounding conditions.
- **Accepted residual (tracked):** the unpinned unlink shares #50's no-generation ABA class — unchanged in kind, documented at each site; when a replace beats `copy_into`'s republish CAS the merge candidate loses its remaining survivors (`Err(RelinkFailure)` discarded) — already reachable via `cas`, now also via `insert`; acceptable-for-a-cache, noted here.


brayniac:
Closing in favor of #55, which fixes the same deadlock (#49) by a different and better route.

**Why #55's approach wins.** Both branches fork from 241b07d and rewrite the same lines of `insert`'s replace arm and `replace_at`, with opposite strategies:

- This PR: never wait — unpinned `cas_location_at` with the decrement skipped.
- #55 (its #51 fix): roll the reservation back and restart. Same-segment (`old_seg_id == new_seg`) rolls back immediately — dropping the `WriterPin` is precisely what unblocks the waiting drain — and re-reserves in a fresh tail; cross-segment gets a bounded spin, then the same rollback to break cross-thread cycles.

In the same-segment case — a re-set of a recently written key, the common one — this PR publishes the new value *into the segment being drained*, so the drain sweeps it moments later: `insert` returns `Ok` and the key is immediately gone. Legal for a cache, but a read-your-writes surprise, and the regression test here never caught it because it exercises the cross-segment shape. #55 handles that case explicitly and preserves exact pinned-unlink accounting, so it needs none of the gauge-settling machinery this PR grew.

**And this approach cannot be rebased forward.** #55's `cas` re-verifies its token under a `NumericVersionGuard`, and `lock_numeric_version` *stores* an odd version word into the item's segment bytes through a raw pointer — a write, not a read. With no pin obtainable on a `Draining` segment, that would be a use-after-recycle write. This PR was sound on its own base (where the cas publish was a pure hashtable slot CAS with the token checked by an earlier pinned read), but not on #55's. The invariant, recorded in #56: *a publish superseding an item may proceed unpinned iff it touches no segment bytes; a publish that must re-verify or freeze item state requires a remover pin and fails safe with `Exists`.*

**What survives from this work:**

- The `s3fifo_promote_from` gauge bug it found — a genuine pre-existing under-count on main (`copy_into` compensates its `remove_item_at` with a batch `ITEM_CURRENT.add`; the promote path does not) — will be re-landed on top of #55, along with an `ITEM_CURRENT`-returns-to-zero-after-drain-storm regression test.
- The insert-side never-wait design is recorded in #56 as the stronger liveness property for a possible future follow-up (#55 terminates provided drains progress; each restart re-runs `reserve_and_define`, so hot-key overwrites pay write-tail amplification).
- The gauge-settling change here is *not* being ported: #55 already corrects the same drift in `reset_write_stats`, idempotently (it zeroes the counters), which the review of this reconciliation established is the better site.

The two deterministic wedge tests here don't transfer — they park a segment permanently in `Draining`/`Relinking`, which is benign under never-wait but would spin #55's rollback-restart loop; a variant that finalizes the drain from a background thread will cover #55's actual liveness property instead.

PR #60, opened 2026-08-19
segcache: guard read-path key verification against stale-location ABA (false-absent lookups)

brayniac:
## Summary

`SegmentsVerifier::verify` (`hashtable/mod.rs:65`) compares key bytes by reading raw
segment memory at the location it is handed, holding **no pin and no generation tag**.
Five read-side bucket-search helpers called `verify` and, on `false`, concluded
"different key" and moved on — ending the scan and returning a **false absent for a
live key**.

The write path already guards exactly this (`try_replace_existing`, added in #46). This
PR mirrors that guard into the read paths and routes all six production `verify` sites
through a single `verify_slot` helper that owns the guard and the invariant it rests on.

## Mechanism

1. Reader loads slot → `location = (S, off)`.
2. Reader stalls (preemption). A merge drain relocates the item via `cas_location`
   (`segments/segment.rs`), `S` is finalized, recycled, and rewritten by another writer.
3. Reader resumes; `verify` reads the **new occupant's** key at `(S, off)` → mismatch →
   the search loop concludes "different key", the scan ends → `lookup` returns `None`.
4. `get_pinned` (`segcache.rs`) returns at `self.hashtable.lookup(key, &verifier)?`
   — **before** any retry logic, so #55's unbounded pin-failure retry never runs.

This predates #55: it lives in the multi-choice hashtable, not the pin protocol. It only
became reachable with concurrent shared-cache use. Published 0.4.0 / 0.4.1 / 0.4.2 are
all affected.

## Sites

All six production `verify` call sites in `hashtable/table.rs`, now all going through
`verify_slot`:

| # | helper | reached via | guarded |
|---|---|---|---|
| 1 | `search_bucket_for_get` | `lookup` → `get` | yes |
| 2 | `search_bucket_no_freq` | `lookup_no_freq_update` → `get_no_freq_incr`, post-pin re-validation | yes |
| 3 | `search_bucket_no_freq_slot` | `lookup_slot` → `cas`, `numeric_update`, `try_into_numeric` | yes |
| 4 | `search_bucket_exists` | `contains` | yes |
| 5 | `search_bucket_for_freq` | `get_frequency` | yes |
| 6 | `try_replace_existing` | `insert` — **already guarded (#46)** | refactored onto the shared helper |

None left unguarded. Sites 4 and 5 have no production caller today (trait surface,
exercised by unit tests), and the question was raised whether guarding them turns a
cheap probe into a spin under churn. It does not: the guard is a **same-slot re-read**,
so each retry requires another thread to have committed a CAS on that slot — it is
bounded by real progress, not a spin on a stable word. And both are exact-answer queries
by contract (`contains` → "is this key present", `get_frequency` → "what is its
counter"), so a false `false` / false `None` is the identical bug waiting for the first
production caller. Uniform guarding is the right call here; no site is better served by
a looser or tighter bound.

## Soundness — the STALE-LOCATION INVARIANT

Stated once, in `verify_slot`'s doc comment, referenced from all six sites. Slot word
unchanged across the verify ⇒ the compared bytes were this entry's throughout ⇒ a
`false` really is a different key. That rests on two clauses:

- **(a) A published entry's location cannot have been recycled while its slot still
  points at it.** A drain unlinks or ghosts *every* entry it drains
  (`try_unlink_in_bucket` / `try_to_ghost_in_bucket`) before the source segment may be
  finalized and reused. #46's same-slot CAS retry is what makes that reliable — a lost
  race there would leave a dangling entry and break this clause.
- **(b) Relocation mutates the slot in place** (`cas_location` / `cas_location_at`), so
  it always surfaces as a changed word.

If either clause stops holding, all six sites become unsound together. Routing them
through one function is deliberate: a change that breaks the invariant is discoverable
rather than silently unsound at six scattered sites.

**The invariant is also checked, not just documented.** Under `debug_assertions`, when a
verify fails and the re-read shows the slot unchanged, `verify_slot` asserts a second
verify agrees. A published item's key bytes are immutable in place (delete's
`set_deleted` runs under a remover pin and touches only the header; numeric updates
rewrite value/CRC, never the key), so disagreement means clause (a) or (b) has broken
and the "unchanged slot" conclusion is unsound. That converts a rare *outcome* into a
frequently-exercised *precondition*: every debug-build verify miss now tests it. It is
`#[cfg]`-gated, so it is compiled out of release entirely.

**Termination.** A `Changed` retry is paid for by another thread's successful `Release`
CAS on that exact slot, so retries are bounded by real system progress. The one
repeatable mutation, a frequency bump, saturates (probabilistic above 16, hard cap 127).
Same bound the existing `cas_location` / `try_unlink_in_bucket` retry loops rely on.

**Miss bookkeeping.** `lookup` bumps ghost frequencies for every candidate bucket on a
miss, so a false absent would also inflate S3-FIFO ghost counters and bias admission for
that key. The retry lives *inside* the per-bucket helper and returns `Some` on success,
so a rescued lookup returns before `lookup` ever reaches
`increment_ghost_freq_in_bucket` — no spurious ghost bump is left behind.

## Scope: this fixes the outcome, not the data race

The fix is detect-and-retry, so **the racing access itself remains**:
`SegmentsVerifier::verify` → `RawItem::key` still races a writer's `ItemHeader::init`.
TSan on this branch produces the **same 7 reports as unfixed main** — expected, and
stated here so nobody reads this PR as having eliminated the race.

What changes is the *observable outcome*: a lookup can no longer turn a live key into a
false absent. Removing the racing access needs generation-tagged locations (#50), which
is the follow-up this PR's residual points at.

## Documented residual

Unchanged from the existing #46 guard: a **byte-identical `packed`** re-published into
the same slot between the two loads reads as unchanged. That needs a full
unlink → recycle → republish carrying the same tag, freq, *and* location value between
two adjacent loads — the same class the `cas_location` retry loops already accept.
Generation-tagged locations are the broader fix, tracked as a follow-up.

## Client-visible impact

Every read-side entry point inherits the false absent:

- `get` / `get_no_freq_incr` — spurious miss, breaking read-your-writes.
- Downstream in pelikan: `add` (which is `get_no_freq_incr().is_some()`) **clobbers a
  live key**; `replace` returns `NOT_STORED` on a live one.
- `delete` returns `false` on a live key.
- `cas` returns `NotFound` where memcached requires `EXISTS`.
- Worst: `try_into_numeric` falls through to `insert(key, initial, ttl)` on a false
  absent — **silently resetting a live counter**.

Measured at ~1.3e-9 per lookup under continuous full-heap turnover (about one CI job in
3000). Rare, but the failure modes include silent clobbering, not merely a legal cache
miss.

## Evidence

Two deterministic layers, making two different claims. Neither is stress volume — a
green run at n=10,000 is weak evidence either way about a ~1.3e-9 event.

### 1. loom model — exhaustive (primary evidence)

`hashtable/` has no `std` sync outside tests, so it is already loom-clean, and
`KeyVerifier` is the boundary the model needs: the existing loom models just stub it with
`AlwaysVerifier`, which cannot express the one thing this bug needs — **a location whose
bytes were rewritten under a reader**. Swapping in a stateful location→key oracle
(`RecyclingOracle`, ~40 lines, zero production change) makes the bug loom-visible.

The model runs a reader (`lookup` / `contains`) against a drain thread that performs the
relocation in production order — copy, publish, recycle — so it never manufactures a
mismatch the real system could not produce: `DST` becomes valid *before* the relink CAS,
`SRC` becomes invalid only *after* it. The key is never deleted, only moved, so `lookup`
must return it in every interleaving.

| model | with fix | guard removed |
|---|---|---|
| `loom_lookup_survives_relocation_and_recycle` | pass (0.16s) | **FAIL (0.06s)** |
| `loom_contains_survives_relocation_and_recycle` | pass | **FAIL** |

Exact pre-fix failure loom reports:

```
thread 'hashtable::table::loom_tests::loom_lookup_survives_relocation_and_recycle'
panicked at crates/segcache/src/hashtable/table.rs:2801:13:
false absent: a relocation + recycle racing the key comparison must not turn a live
key into a miss (STALE-LOCATION INVARIANT)
test result: FAILED. 0 passed; 1 failed; ... finished in 0.06s
```

```
thread 'hashtable::table::loom_tests::loom_contains_survives_relocation_and_recycle'
panicked at crates/segcache/src/hashtable/table.rs:2844:13:
contains must not report a live key absent
```

### 2. Plain deterministic tests — one per read entry point

The loom model covers `lookup` and `contains`; a model per helper would multiply
state-space cost for little added signal. These plain tests cover all five read entry
points cheaply, and read as regression tests rather than as a model. They use the same
insight — `KeyVerifier` is the seam — with a `RelocatingVerifier` that performs the
relocation from inside its own `verify` call. No production hook, no `#[cfg]` in the read
path, no scheduler dependence.

I also checked #59's new `fault-injection` knob per the "reuse an existing seam"
guidance and did not use it: it forces `copy_into`'s relink CAS to *lose*, driving the
`RelinkFailure` abort path — a different window from "a relocation lands mid `verify`".

`hashtable::table::stale_location_tests`, 7 tests, **0.00s**:

| test | with fix | guard removed |
|---|---|---|
| `lookup_survives_relocation_during_verify` | pass | **FAIL** |
| `lookup_no_freq_update_survives_relocation_during_verify` | pass | **FAIL** |
| `lookup_slot_survives_relocation_during_verify` | pass | **FAIL** |
| `contains_survives_relocation_during_verify` | pass | **FAIL** |
| `get_frequency_survives_relocation_during_verify` | pass | **FAIL** |
| `insert_replaces_through_relocation_during_verify` | pass | pass¹ |
| `genuine_tag_collision_still_reports_absent` (control) | pass | pass |

"Guard removed" = `classify_failed_verify` reduced to `SlotVerify::DifferentKey`, which
is exactly the pre-fix semantics (conclude "different key" without re-reading).

¹ `insert` re-checks for an existing entry under the stripe lock, and by then the
relocation has settled, so the second pass finds it. That test pins the outcome (replace,
never a duplicate); it is not a reproducer, and its doc comment says so.

The control test matters, and it took a correction to earn its place. It originally used
unrelated keys (`b"present"` / `b"absent"`); with a 12-bit tag the SIMD mask screens such
a probe out *before* `verify` runs, so instrumentation showed it made **zero** calls into
`classify_failed_verify` — it could not have failed if the guard had degraded into
retry-on-every-mismatch, which is the one regression it exists to catch. It now
constructs a genuine collision (two keys sharing a tag **and** their first candidate
bucket, searched against the live `RandomState`-seeded table) and asserts that
precondition inline, so it cannot silently go vacuous again.

Verified in both directions:

- **Coverage:** instrumented, the new test makes **5** calls into
  `classify_failed_verify` — one per read entry point — all resolving `DifferentKey`.
  The old form measured **0** on the same instrumentation.
- **Catches the regression:** with `classify_failed_verify` neutered to return
  `Changed` unconditionally, the new test **hangs** (the slot never changes, so the retry
  never terminates) and is killed at a 15s timeout, while the old form still **passed**
  under the identical neutering. The regression surfaces as non-termination rather than
  a wrong answer, so it is caught by CI timeout — worth knowing if it ever fires.

### 3. Instrumentation across the whole lib suite (positive evidence)

The red/green tests show the failure mode is *caught when it occurs*. This shows the
invariant actually *holds under real traffic* — the stronger claim, and the thing to
point at if this is ever questioned.

Both arms of the guard were instrumented and the full `--lib` suite run under
`--test-threads=1` (three runs):

| outcome | count |
|---|---|
| `classify_failed_verify` calls | 12,417 – 12,423 |
| → `DifferentKey` (slot unchanged) | 12,409 – 12,415 |
| → `Changed` (slot moved under the verify) | 8 |
| `debug_assertions` precondition evaluated | **12,409 – 12,415 (100% of `DifferentKey`)** |
| precondition **failures** | **0** |

Two things worth drawing out:

1. The `Changed` arm is **not dead code** — it fires 8 times per suite run, so the race
   this PR fixes is genuinely reachable in the existing tests, not a theoretical
   construct.
2. The precondition count exactly equals the `DifferentKey` count on every run, i.e. the
   invariant is checked on *every single* "slot unchanged" conclusion the suite makes,
   and it never once disagreed. That is ~12,400 independent confirmations per run that
   an unchanged slot really does imply stable key bytes.

## Stress (smoke check only, not the primary evidence)

Before switching to the deterministic seam, `pin_failure_tests::merge_churn_no_false_miss_no_resurrection`
under `--features debug` at 24-way parallelism on a 32-core machine:

| | runs | failures |
|---|---|---|
| pre-fix (`e3d83c3`) | 14,400 | **4** |
| post-fix | 34,400 | **0** |

Pre-fix failure, verbatim:

```
thread '<unnamed>' panicked at crates/segcache/src/pin_failure_tests.rs:638:21:
hot key h000003 reappeared after a miss: false miss during a merge drain
```

Reported for completeness. The deterministic tests above are the evidence.

## Performance

The guard only runs when `verify` returns `false`, which needs a 12-bit tag collision
(~1/4096 per examined slot) or the race itself — so it should be off the hot path. It
was measured rather than assumed, on the `get` criterion benchmark, branch vs. main:

Criterion, quiet machine, `--baseline` against main:

| bench | main | this branch | change |
|---|---|---|---|
| `get/1b/0b` | 17.750 ns | 17.068 ns | **-4.4%** (p<0.05) |
| `get/255b/0b` | 33.630 ns | see note | — |

Same-binary control (main measured twice against its own baseline): **-0.34% / -0.32%**,
so run-to-run drift on this machine is well under +/-0.5%.

_Note:_ a concurrent heavy job on the shared machine spoiled the `get/255b/0b`
re-measurement of the final form (its confidence interval widened to 34.8-44.7 ns). A
load-robust alternating A/B (both binaries interleaved, min-of-8, so load inflates both
arms equally) showed **no regression**: `get/1b/0b` 37.2 ns (branch) vs 40.2 ns (main),
`get/255b/0b` 72.8 ns vs 72.3 ns — ahead on the small key, a 0.7% tie on the large one.
Absolute values are inflated ~2x by the load; only the comparison is meaningful. I have no measurement
showing a regression in the final form; the -4.4% on `get/1b/0b` is most likely code
layout, and I would not claim it as a real speedup.

The first cut folded the re-load inline into `verify_slot` and cost a reproducible
**+1.1% / +1.9%** (1b / 255b) against a ±0.4% same-binary control — the extra code in the
inlined hot body, not the load itself. Splitting the failure half into a
`#[cold] #[inline(never)]` `classify_failed_verify` removed it. The attributes are
load-bearing; there is a comment on the function saying so.

## Test changes

`merge_churn_no_false_miss_no_resurrection` keeps its assertion — it is sound and it
caught a real bug. Two diagnostic improvements:

- The panic now names both bugs that land there:
  `false miss (drain window or stale-location verify)`.
- It carries how many polls the key took to reappear, so a future flake explains itself:
  an immediate reappearance points at the verify path, a long one at a drain window.

## Unrelated pre-existing flake (not fixed here)

`numeric_concurrency_tests::cas_incr_stress_exact_accounting` fails intermittently with
`counter must stay resident (no eviction pressure)` — a `get` false miss on a hot key
under `cas`/`incr` churn. **It has hit this PR's macOS CI job**, so it is worth being
explicit: it is pre-existing and unrelated to this change.

Evidence — the two builds (this branch vs `b9bac23`) run interleaved on one machine so
both arms see identical load:

| conditions | main | this branch |
|---|---|---|
| serial, 250 runs/arm | 4 | 4 |
| 24-way oversubscribed, 300 runs/arm | 3 | 6 |
| 24-way oversubscribed, 1200 runs/arm | 19 | 18 |
| **total, 1750 runs/arm** | **26 (1.5%)** | **28 (1.6%)** |

Statistically identical. The rate rises sharply with CPU oversubscription, which is why
a 3-core macOS runner hits it far more often than a 32-core dev box — and why it showed
up here rather than on `main`'s single post-merge run.

Root cause is almost certainly `get_pinned`'s bounded-`attempts` giveup: `cas` moves the
key to a new location on every iteration, so the post-pin re-validation can lose the
race `RESERVE_RETRIES` times in a row and return `None` for a key that is very much
alive. A different bug from this one — that path gives up *after* a correct lookup,
whereas this PR fixes the lookup itself returning a wrong answer. Happy to open an issue
and take it separately.


brayniac:
Follow-up for the CI flake noted in the PR body: filed as #65.

Instrumenting every `None` exit in `get_pinned` confirms it is **not** this PR's path — `lookup_none=0` across all observed failures, i.e. the hashtable lookup never returns `None`. It is `get_pinned`'s bounded post-pin re-validation (`RESERVE_RETRIES = 3`) giving up after a *correct* lookup, which is a second, narrower false-absent path (`get`/`get_no_freq_incr` only). Details, rate, and a soundness assessment of the obvious fix are in #65.

brayniac:
**Review sign-off (independent adversarial pass).** Approving in substance — GitHub won't take a formal approval since both work streams push as the same account. No blocking correctness or concurrency defect. I reviewed this as the author of the write-path guard being generalized here.

## Verified rather than taken on trust

**The STALE-LOCATION INVARIANT's two clauses hold.** Every path returning a segment to a queue passes through `Segment::clear` first — `finalize_drained`→`recycle`, `condemn`, `drain_chain` (expire and flush), and `remove_at`'s empty-free via `clear_segment`. The one queue-return that skips a sweep, `release_unused`, is reached only by chain-extension election losers holding a never-published `Reserved` segment; merge destinations always exit via `publish_dest_sealed`, including on `RelinkFailure`. The sweep is exclusive: `claim_for_drain` wins the `Sealed→Draining` CAS then drains `active_writers`/`active_removers`, while publish happens under the `WriterPin`. For clause (b), the complete set of occupied-slot writers was enumerated — every relocation is a single-slot CAS on the entry's own slot, and with no rehash and no resize nothing moves an entry between slots or buckets.

**The debug assert's rationale is stronger than stated.** `set_deleted` only ORs a flag bit and `klen` is a separate field — but more decisively, `Segcache::delete` calls `hashtable.remove` *before* `set_deleted`, so the entry is already unlinked. `numeric_update` writes value/version/CRC only; `define()`, the sole writer of key bytes, runs pre-publish under a `WriterPin`.

**`DifferentKey` under ABA needs two full recycle+refill cycles**, not one: for the reload to match bit-for-bit, the segment must be recycled and refilled to the *same* offset with the *original* key, and the fresh insert must claim the *same* slot. (The `freq=1` field does line up, since `try_claim_new_slot` publishes fresh entries at 1 — the one field adding no improbability.) Strictly harder than the write-side `cas_location` ABA already accepted here. The PR's "accepted residual; generation-tagged locations are the broader fix" wording is accurate and if anything understates the margin.

**Termination.** `Changed` requires the word to differ from the `packed` loaded at the top of the same iteration, so every retry is paid for by another thread's completed Release CAS. Reader freq bumps can't spin it — a bump leaves the location unchanged, so `verify` returns `Match` and the loop exits; `Changed` is reachable only after a verify *failure*, and `try_update_freq` caps at 127. The SIMD mask is consumed once per slot before the inner loop, which re-checks `packed == 0`/`is_ghost`/tag every pass, so a stale mask bit resolves to `break` — no slot skipped, none visited twice.

**Behavior preserved at all six sites**, diffed individually against `b9bac23`. The `try_replace_existing` site is byte-equivalent to the old inline guard, including `allow_deleted: true` and the untouched F4 CAS-failure `continue`. The read sites' old outer-`continue` became an inner-loop `break` falling to the same outer iteration — identical control flow. No production `verify` remains outside `verify_slot`.

**The tests are not vacuous.** Neutering the guard (`Changed => continue` → `break`) turns all five read entry points red and both loom models red, while the two negative tests stay green. `RecyclingOracle` is a consistent oracle sequencing `dst_valid=1` before the relink CAS and `src_valid=0` after — copy→publish→recycle — driving the real `lookup`/`contains`. Instrumenting both arms over the lib suite: **12,410 `DifferentKey` and 10 `Changed`** outcomes, with the debug precondition live on every one of the 12,410 and never firing. That's evidence *for* the invariant, not just against the failure mode.

Gates re-run independently: full suite, `--features debug` (130+19), loom 24/24, clippy `--all-features -D warnings`, fmt — all clean. The `#[cold] #[inline(never)]` split is load-bearing and correctly commented: on a hit the inner loop runs once and `verify_slot` is just the comparison, so the hot path structurally doesn't pay, corroborating the reported A/B.

## Non-blocking follow-ups

1. **`genuine_key_mismatch_still_reports_absent` doesn't test what its doc claims.** It says it guards against "the fix turning every mismatch into a retry", but it inserts `b"present"` and looks up `b"absent"` — with a 12-bit tag the SIMD mask almost never matches, and instrumentation shows **0 calls** into `classify_failed_verify`. It's a smoke test, not a `DifferentKey` guard. Craft colliding keys (as `test_replace_across_buckets_no_duplicate` already does) or seed a `MockVerifier` entry that forces the mismatch.
2. `search_bucket_for_freq`'s rationale comment says a false `None` "would feed the eviction policy a wrong frequency" — eviction uses `get_item_frequency` (location-keyed, no `verify`), and `Hashtable::get_frequency` has no production callers. Guarding it is still right; the stated reason isn't.
3. Worth a sentence that `SegmentsVerifier::verify` ignores `allow_deleted` entirely, so the `true`/`false` split between write and read sites is inert in production and observable only to `MockVerifier`.
4. Optional: `debug_assert!(!verify(..) || slot.load(Acquire) != packed, ..)` narrows a theoretical false positive where the slot is unlinked/recycled/refilled between the word comparison and the second verify.

(1) is worth doing — a test whose doc claims coverage it doesn't have is worse than no test. The rest are comment-only. None gate the release: fold them in here if cheap, or track them.

Issue #58, opened 2026-08-19
segcache: ITEM_DEAD/ITEM_DEAD_BYTES are monotone counters labelled as gauges; condemned segments skip residue reconciliation


brayniac:
Two related item-accounting follow-ups, both surfaced by the review of the gauge-accounting PR (which fixes the `ITEM_CURRENT`/`ITEM_CURRENT_BYTES` side and deliberately leaves these).

## 1. `ITEM_DEAD` / `ITEM_DEAD_BYTES` never decrease

There is no decrement site anywhere in the crate — every write is an `increment()`/`add()` (`segment.rs` `remove_item_at`, and `header.rs` `reset_write_stats` as of the gauge PR). But `metrics.rs` describes them as *"current number of dead items"* / *"current number of dead bytes"*. They are cumulative counters wearing a gauge's label, and dead bytes really are reclaimed at recycle, so the reported value diverges from reality without bound.

The gauge PR corrects the two description strings to say cumulative — an honest label for current behavior, not a fix. The real question is which semantics are wanted:

- **Gauge (as documented):** decrement at recycle, alongside the existing residue reconciliation in `reset_write_stats`.
- **Cumulative (as implemented):** keep as-is, and additionally decide whether a *relocation* should count as a death at all. It currently does: `copy_into` / `s3fifo_promote_from` call `remove_item_at` (which bumps `ITEM_DEAD`) while being deliberately `ITEM_CURRENT`-neutral, so merges inflate "dead" without anything having died.

Note `crates/segcache/tests/item_gauges.rs` asserts `item_dead == item_allocate + item_relink`, which encodes the current behavior including relocations-count-as-deaths. Changing the semantics means updating that assertion — the comment there points here.

## 2. Condemned segments skip residue reconciliation

`reset_write_stats` — which subtracts the unpinned-unlink residue from the global gauges — runs from `recycle` and from `try_reserve`. The **condemned** path reaches neither: `condemn` → the last reader's `SegmentGuard::drop` → `free_queue.push(...)` returns the segment to `Free` with its residue intact, reconciled only if and when it is later re-reserved.

[...]

PR #63, opened 2026-08-19
segcache: fix reader-pin reclamation of condemned segments

mux:
# segcache: a condemned segment can be freed while a reader pin is live

**Disclaimer.** This report was written by Claude on my behalf, in the interest of time. We understand segcache is still under active development, so you may already know about all of this, and it is entirely possible we have misread something and one or more of these findings is unfounded. Everything below was derived by reading the code at `e3d83c3`; we have not reproduced any of it in Rust. The change proposed here builds and passes your suite. The "Defences we checked" section at the end lists what we went looking for and ruled out, so you can see the shape of the analysis rather than take it on trust.

These are distinct from the six races fixed in 51b154b (#55), which addressed the pin-*failure* protocol (drain false-absence, delete resurrection, replace-vs-drain deadlock). The condemned-release path is untouched by it: all three sites below are unchanged on `e3d83c3`.

Related but, we believe, distinct: #50 already records that a stale `Location` can pin a *new incarnation* because `Location` carries no generation, on `Segcache::insert`'s fresh-key arm via `try_pin_remover`. Finding 2 below is the same defect class -- a token with no lifetime identity -- applied to the metadata word on the *release* side, and findings 1 and 3 are about who frees a condemned segment and whether its reader count survives. If you consider any of this subsumed by #50 or by #58's condemned-segment accounting work, say so and we will drop it.

## Root cause

`State::AwaitingRelease` is readable (`segments/state.rs`), justified by:

> new reads cannot arrive because the hashtable is fully drained before a segment is condemned

Draining stops new *lookups* from routing to the segment, but the lookup and the pin are two separate steps in `get_pinned`: `hashtable.lookup` returns a `location`, and only then does `acquire_item_at` pin the header. A reader stalled in that gap calls `try_acquire_reader` after the condemn, and AwaitingRelease being readable means the pin succeeds. So a new *pin* can arrive after a segment is condemned even though no new *lookup* can. Once the reader count can return to non-zero after reaching zero, no reference-count handoff is sound.

## 1. Free under a live pin

1. A holds the last pin. `SegmentGuard::drop` calls `release_reader_for_guard()`, which returns 1; the count is now 0.
2. B, whose `hashtable.lookup` already returned this location and which stalled before pinning, now calls `try_acquire_reader`: AwaitingRelease is readable, so `fetch_add`, re-check still readable, **pin succeeds**.
3. A's `try_release_condemned()` sees AwaitingRelease, CASes to Free, and pushes the id to the free queue.

The segment is now in the free queue with a live pin on it, so it can be reserved and overwritten while B still holds that pin. B will not *serve* stale bytes -- `get_pinned` re-validates the location after pinning, and the entry is gone from the hashtable -- but B's eventual decrement lands on the *next* incarnation's counter. Neither `recycle()` nor `try_reserve()` resets `ref_count`, so that incarnation carries a phantom count for the rest of its life: `finalize_drained` never sees 0 for it, so it can never be recycled again, only condemned.

Note that `recycle()` already asserts this exact invariant:

```rust
debug_assert_eq!(
    self.headers[id_idx].ref_count(),
    0,
    "freed a segment pinned by readers"
);
```

The condemned-release path does not go through `recycle()` -- the guard pushes to the free queue directly and `condemn` uses `return_segment` -- so neither carries that assertion.

## 2. Stale release CAS (independent of finding 1)

`try_release_condemned` CASes the whole metadata word from the value it observed. For a condemned segment that word is the constant `{AwaitingRelease, prev: None, next: None}`, identical for every use of every segment, so the token carries no lifetime identity.

1. T1 (guard drop) loads `current`, then stalls.
2. The condemner's race-fix frees the segment; it is recycled, refilled, evicted and condemned again, now pinned by T3.
3. T1 resumes, the word matches, the CAS succeeds, and T1 pushes the id.

Incarnation N+1 is in the free queue under T3's live pin, and T3's own drop finds the state already Free and declines, so the handoff is stolen rather than duplicated. `generation` exists but is not in this word. It is consumed in exactly two places: `cas.rs`, for item CAS tokens, and `ttl_bucket.rs`, as a stale-observation ABA guard (`segments.header(tail_id).generation() != gen`). That second use is the same pattern item 3 of the fix applies here, so the approach has precedent in the tree rather than being something we invented.

## 3. Acquire backout can strand a condemned segment (live today)

`try_acquire_reader`'s re-check failure backs out with a plain `fetch_sub(Release)`. A reader that pinned during the Sealed to Draining transition takes this path. If `finalize_drained` observed that transient pin it condemns and defers reclamation to "the last reader", who never checks the state, and the plain decrement performs no handoff. Nothing anywhere sweeps AwaitingRelease segments, so the segment is stranded permanently. This one does not depend on the readability of AwaitingRelease and is reachable today.

The same unaccounted transient pin also reaches the *recycle* arm. In `TtlBucket::expire`, a `header_ref_count_seqcst() == 0` check is followed by `segments.recycle(seg_id)`; a backing-out acquire whose `fetch_add` lands after that check but before `recycle` gives a non-zero count inside `recycle`, whose `debug_assert_eq!` then fires. So there is a testable prediction attached to this finding: if anyone has ever seen `"freed a segment pinned by readers"` trip spuriously in a debug build, this is a candidate cause. In a release build the segment is instead reused while the transient pin is outstanding, and the backout's decrement underflows the new incarnation's counter.

## On the guard doc

`guard.rs` proves "exactly-one-free" from CAS uniqueness. That is true, but the hazard is *one* free under a live pin, which CAS uniqueness does not address.

## The fix

1. Drop `AwaitingRelease` from `is_readable()`, so no pin can *succeed* after condemn and `prev == 1` means no reader remains. The count is not strictly monotonic even then -- a failing acquire still bumps it transiently between its `fetch_add` and its re-check -- which is what item 2 exists to make safe. We could not identify an observable cost to this change: a reader already holding a pin never re-checks `is_readable`, so it completes either way, and a reader acquiring after condemn fails the post-pin re-validation regardless, because `clear` removed or relocated its entry before the condemn. The difference is only whether it fails at the pin or one step later.
2. Make the acquire backout use the same SeqCst decrement plus condemned handoff as a guard drop, so a transient pin the condemner saw still gets freed. `try_acquire_reader` returns a new `AcquireOutcome` so `acquire_item_at` can return the segment; it has `&Segments`, which the guard does not.
3. `cas_condemn()` stamps the low byte of `generation` into the spare high byte of the metadata word (`[8 unused][8 state][24 prev][24 next]`), making the release token unique to one use of the segment. Eight bits aliases every 256 uses; widening means narrowing `prev`/`next`.

**This change does not cover the recycle-arm variant** noted under finding 3. There the segment is recycled rather than condemned, so the backout's handoff finds no AwaitingRelease state to claim and the stray decrement still lands on the next incarnation. Closing that needs a choice on the recycle path -- wait out transient pins, or restructure the acquire so it does not increment before it knows the state is readable -- and we did not want to make that call on your behalf.

`cargo test -p segcache`: 138 passed, 0 failed. `cargo check -p segcache --features loom --tests`: clean.

## Defences we checked

Things that could have made the above wrong, and what we found:

- **Is the lookup-before-pin window real?** Yes, but it is narrow on the reader's side, and we want to be straight about that. B's gap is the few instructions between `hashtable.lookup` returning and `try_acquire_reader` running (`unpack_location`, `NonZeroU32::new`, the call, a bounds assert, an array index). To land after the condemn, B must be descheduled or stalled there while the drain removes B's entry, finishes the rest of `Segment::clear` (a loop doing `get_item_frequency` plus `remove` for every item in the segment), reads `ref_count`, and condemns. So it needs a reader stall coinciding with a drain, not an ordinary interleaving. Findings 2 and 3 need a stall too, and finding 2 needs by far the longest one: T1 must be stalled between two adjacent instructions while that *specific* segment id is freed, re-reserved, refilled, evicted and condemned again, and the free queue is FIFO so the id does not come back quickly. That last point cuts both ways, though: how long the stall must be scales with the number of segments, so a small cache recycles ids fast enough to make this materially more reachable than it looks on a large one. We would call all three latent rather than observed, and that is why we have not tried to build a deterministic repro.
- **Post-pin re-validation.** `get_pinned` re-looks-up the location after pinning, so a reader cannot serve bytes from a reused segment. This narrows finding 1: the harm is the free-under-pin and the counter corruption, not stale data. Stated that way above.
- **Does the fix livelock `get_pinned`?** Its pin-failure arm retries unbounded and is deliberately not counted against `attempts`, so making a state unreadable is dangerous. It is safe here because `finalize_drained` runs `segment.clear(hashtable, expire)` *before* `condemn`, so no hashtable entry resolves to a condemned segment and the top-of-loop lookup returns `None` instead of spinning.
- **Is anything sweeping AwaitingRelease?** No. The only occurrences are the state check, the condemn transition, and tests, so finding 3 strands permanently rather than transiently.
- **Are there other callers we missed?** `try_release_condemned` has exactly two production callers, the guard drop and `condemn`'s race-fix, so the analysis covers the whole release path.
- **Refutation attempt: does anything reset `ref_count` on reuse?** This would have killed finding 1's stated consequence. It does not. Every write to `ref_count` in the crate is the acquire's `fetch_add`, `release_reader`'s and `release_reader_for_guard`'s `fetch_sub`, and a `store(1)` inside a loom test. No reset on any reuse path.
- **Refutation attempt: does `claim_for_drain` wait for readers?** This would have killed finding 3, since `finalize_drained` could never observe a transient reader pin. It waits on `active_writers` and `active_removers` only; the reader count is deliberately handled by the Dekker pair instead.
- **Refutation attempt: does borrow checking serialize drains against reads?** If the drain path needed `&mut Segments`, safe Rust would make all three unreachable. It does not: `Segcache::get`, `Segcache::expire`, `TtlBucket::expire`, `claim_for_drain`, `finalize_drained` and `condemn` all take `&self`, so reads and drains run concurrently on a shared reference.
- **Did #55 already fix any of this?** Its six races are in the pin-failure protocol. All three sites here are byte-identical on `e3d83c3`, and `try_reserve` still does not reset `ref_count` despite that commit touching it.

## What we saw in C

Our port already makes the condemned state non-readable, so finding 1 does not apply there, but it still needed a reader-count recheck at the last unpin and a generation in the state word. A stress test holding long read pins on hot multi-segment entries under constant eviction reproduced readers seeing *other keys'* entries, 2 of 3 runs at 32 threads, before those two fixes.


brayniac:
Thanks for this — the analysis is considerably better than the disclaimer suggests, and I want to lead with the verdict: **all three defects are real**, and I reproduced two of them deterministically against current `main` (c541316), not just the e3d83c3 you read. Your patch passes the full gate locally, both as submitted and cherry-picked onto `main` (it applies cleanly).

Your note that the C port needed *both* a reader-count recheck at the last unpin *and* a generation in the state word matches what I concluded independently — including that those two parts have to ship together. More on that below, because it's the main decision here.

## Verification, per defect

**Defect 1 — free under a live pin: confirmed, reproduced deterministically.** No new seam was needed: `SegmentGuard::drop` is `release_reader_for_guard()` followed by `try_release_condemned()`, so `mem::forget` on the `Item` plus calling those two halves by hand puts a stalled reader exactly in the gap. All three of your claims land — the post-condemn pin succeeds, the segment reaches `Free` with `ref_count == 1`, and `try_reserve` does not reset it, so the next incarnation starts with a phantom count.

One narrowing, since you asked to be corrected: **the phantom count is transient, not lifelong.** The late reader's `fetch_add` and eventual `fetch_sub` land on the same header, so the arithmetic rebalances when it drops; the damage is bounded by that stall, during which the new incarnation gets spuriously condemned rather than recycled — itself a legal path. I also went looking for a route from defect 1 *alone* to reader-visible corruption and could not construct one at `main`: `Segment::clear` empties the hashtable of the condemned segment before `condemn` runs, and all four `acquire_item_at` callers re-validate the location after pinning, so the late reader's re-validation fails and it never dereferences the item. The narrow exception is `clear`'s `hashtable.remove(...) == false` branch. None of that lowers the priority — a free under a live pin breaks a contract other code is entitled to rely on, and defect 2 turns the same broken invariant into unambiguous corruption.

**Defect 2 — stale release CAS with no lifetime identity: confirmed, and I'd rank it above defect 1**, higher than your own ordering. It is the only one of the three that can free a segment out from under a reader holding a **validated** `Item`. Defect 1's free is gated on `prev == 1`, so no validated holder exists at that instant; your stale CAS is gated on nothing, and can win against a later incarnation that is legitimately `AwaitingRelease` with live readers whose re-validation already passed. Their `Item::value()` then reads reused memory.

**Defect 3 — acquire backout strands a condemned segment: confirmed, but your reachability claim is wrong in a way that strengthens your case.** You say it is live today and independent of `AwaitingRelease` being readable. Not for the condemn arm: my first repro attempt put the drain between the acquire's `fetch_add` and its re-check, and the acquire *succeeded*, because at `main` the re-check sees `AwaitingRelease`, which is readable — that is defect 1's mechanism, and the resulting real pin hands off correctly. It **is** reachable at `main`, but only through a much tighter window (descheduled between the re-check load and its `fetch_sub`, two adjacent instructions, and staying descheduled across `Segment::clear` plus `condemn`), which I confirmed with a two-phase interposition seam. Crucially it becomes **materially more reachable once your fix 1 lands**, since the re-check then sees `AwaitingRelease` as non-readable too. So fix 3 is not garnish on fix 1 — it is required by it. Nothing anywhere sweeps `AwaitingRelease`, so the consequence is a permanent segment leak.

## On the shape of the fix

Removing `AwaitingRelease` from `is_readable` is the right move, and I want to be explicit about why, because I went in expecting to argue it merely narrowed the window. It doesn't tighten the reader side — it changes what the count *means*. Afterwards, within an incarnation `AwaitingRelease` goes only to `Free`, so no acquirer can ever return `Acquired` again; the count can still blip 0→1→0, but every such pin is doomed by construction and backs out without constructing a `RawItem`. So `prev == 1` becomes stable in the only sense that matters. Worth noting for anyone reading later: **re-reading `ref_count` before the CAS would not have closed this** — that is still check-then-act across two atomics.

**Decision on your generation tag: keep it in this PR.** We have an open issue (#50) for generation-tagged locations, and the tidy-looking move would be to defer your tag to it. That would be wrong: cross-incarnation stability is exactly what the tag provides, so fix 1 without it leaves a live use-after-free — a thread observes `prev == 1` on incarnation N, stalls, N is freed, recycled, refilled and condemned again as N+1 with live validated readers, and an untagged CAS matches and frees N+1 under them. If #50 later widens 8 bits to something better, that is strictly additive.

## Requests

1. **Metrics on the `ReleaseCondemned` arm.** It calls `return_segment` without incrementing `SEGMENT_RETURN`/`SEGMENT_FREE`, which every other free site does. `SEGMENT_FREE` is a Gauge that `reserve_free` decrements, so a missing increment is permanent downward drift and eventually underflow — the same class of bug we fixed in #59.
2. **Stop the condemn tag being silently erased.** `Metadata::pack()` writes zeros into bits 63..56, so `Metadata` does not round-trip the word and any read-modify-write through it drops the tag. `update_links` is exactly such an RMW and is reachable against a currently-`AwaitingRelease` segment, because `condemn` splices neighbours after its own CAS and `recycle`'s unlink does the same — so two adjacent chain segments drained by two evictors means one zeroes the other's tag, restoring the untagged constant. Please fix it by **carrying the tag inside `Metadata` so pack/unpack round-trip it**, rather than preserving the high byte at each call site: preserving bits at call sites is a rule every future writer of that word has to remember, and this bug is that rule already being forgotten once, in the change that introduces it.
3. **A few doc updates the change leaves stale**, most importantly `try_pin_writer`'s backout comment, which says it mirrors `try_acquire_reader`'s backout and justifies `Release` with reasoning this PR disproves for readers. Writers and removers genuinely are different — `claim_for_drain` *waits* on `active_writers`/`active_removers` rather than deferring a handoff — and it would be good to say that instead. Also the two packed-layout diagrams still marking bits 63..56 unused, `guard.rs`'s SAFETY contract still saying `try_acquire_reader` returns `true`, and its module doc describing two racers where there are now three.

## Offered, not imposed

I have deterministic reproductions for defects 1 and 3 working (the defect-3 one needs a small `#[cfg(test)]` two-phase hook in `try_acquire_reader`, ~15 lines). They're yours if you want them — say the word and I'll post them as patches for you to apply or reshape. You reported three real defects from a code read without a Rust repro; the tests should come to you rather than your patch being rewritten underneath you.

Two smaller notes, neither blocking: `AcquireOutcome` is `pub` where the neighbouring `SegmentPool` is `pub(crate)`; and `Policy::RandomFifo` now skips condemned segments via `is_readable()` — I think an improvement, but currently unremarked. Your deliberately-left-open recycle arm also hits `finalize_drained` as well as `TtlBucket::expire`, and its real cost is a spurious `debug_assert` rather than a data hazard, so I agree it's a follow-up and shouldn't block this.

CI is now enabled on the branch. Thanks again — this is a genuinely valuable report.

mux:
Thank you for the review, Brian. I believe I have now addressed your 3 points and the visibility note. I have also added a case to the existing `metadata_roundtrip` test to ensure the tag survives the round trip. The commit log now mentions the `Policy::RandomFifo` change. 

You're right about the recycle arm and my note overstated it: the backout's decrement balances the same reader's own increment, so there's no underflow — a spurious `debug_assert` and a brief phantom count, and it reaches `finalize_drained` as well as `TtlBucket::expire`.

I am happy to incorporate your reproductions if you send them my way.

brayniac:
Thanks — that was fast. A full review of the delta is running; three practical things in the meantime, none of which need you to wait on me.

## 1. Lint is red on `5568ad9`, and it's trivial

`cargo fmt --all --check` wants the `pack()` expression on one line in `crates/segcache/src/segments/state.rs:177`:

```rust
((self.tag as u64) << Self::TAG_SHIFT) | ((self.state as u64) << 48) | (prev << 24) | next
```

That's the only fmt diff on the branch. Running `cargo fmt --all` locally will fix it.

## 2. The defect-1 reproduction, as offered

Pushed to **[`brayniac/cache-rs:repro-for-63`](https://github.com/brayniac/cache-rs/tree/repro-for-63)** — one commit on top of your `5568ad9`, so it should cherry-pick cleanly:

```
git remote add brayniac https://github.com/brayniac/cache-rs
git fetch brayniac repro-for-63
git cherry-pick brayniac/repro-for-63
```

`condemned_segment_is_freed_under_a_live_reader_pin` needs no new seam: `SegmentGuard::drop` is `release_reader_for_guard()` then `try_release_condemned()`, so `mem::forget` on the `Item` plus calling those two halves by hand puts a stalled reader exactly in the gap. It's red at `main` (c541316) and green with your fix — I ran it against your head before pushing.

**The defect-3 seam doesn't transfer as-is**, because it hooks `try_acquire_reader` and your fix rewrote that function. It needs adapting to your new shape rather than mechanically applying, and you're better placed to do that than I am — happy to do it instead if you'd rather, just say. It's a `#[cfg(test)]` two-phase hook that parks a reader between the re-check load and its `fetch_sub`.

## 3. You'll need a rebase, and one of the four merges matters

Your branch is based on `b9bac23`; `main` is now `2788fcf`. Of the four merges in between, **#68 rewrote `get_pinned`'s revalidation retry loop** — which is the loop your livelock argument reasons about, so that argument is worth re-checking rather than assuming it carries. The other three (#60 read-path verify guard, #67 loom coverage, #70 benchmarks) shouldn't touch you. I'm having the review check that interaction explicitly and will report either way.

Thanks for the correction on the recycle arm — agreed the decrement balances the same reader's own increment, so it's a spurious `debug_assert` and a brief phantom count rather than an underflow, and it reaches `finalize_drained` as well as `TtlBucket::expire`. Adding the round-trip case to `metadata_roundtrip` was the right instinct; the review will confirm it would actually fail if the tag were dropped, since a round-trip test that passes trivially is a failure mode we've hit twice this week ourselves.

brayniac:
Reviewed `093a278..5568ad9`. **All three asks are done, and done the way they should have been.** Two one-line items left, plus a correction I owe you.

## Verified

**Metrics** (`segments.rs:328-336`) — matches all four other free sites in shape, fires exactly once (the `ReleaseCondemned` variant is only constructible when `prev == 1 && try_release_condemned()` both hold, and that arm never constructs a guard, so it's mutually exclusive with the guard-drop path), and `return_segment` doesn't increment, so there's no double count.

**Tag round-trip** — the structural version, which is what matters: the tag is a real `Metadata` field, `pack` writes 63..56 and `unpack` reads them, and demoting `TAG_SHIFT` from `pub` to private means no call site *can* OR bits in any more. I traced all 7 `pack()` and 6 `unpack()` sites plus the two in `state.rs`: no path zeroes the tag except `AwaitingRelease -> Free`, which should. Also checked the interactions — no `cas_metadata` call uses `expected_state == AwaitingRelease`, so `tag != 0` stays confined to that state; state extraction masks to 8 bits and is disjoint from 63..56; and while the tag joins `Metadata`'s derived `PartialEq`, nothing in the crate compares two `Metadata` values.

**`Policy::RandomFifo`** — agreed it's an improvement, and for a better reason than I gave: `is_readable()` there only picks a segment in order to pick its *TTL bucket*, so excluding condemned segments stops weighting a bucket the segment has already left.

## Two small things

1. **`cargo fmt`** — `pack()`'s expression on one line (`state.rs:177`, 99 cols). That's the only fmt diff on the branch and it's what CI lint is red on.
2. **`Segments::condemn`'s own doc, `segments.rs:921`** still says the state *"remains readable for in-flight pins"*. That clause is now false, and it's the precise claim this PR exists to refute — same class as the `try_pin_writer` comment we did name. We didn't name it, so this is a fair ask rather than something you missed.

## A correction to my earlier review

My justification for asking you to carry the tag in `Metadata` doesn't survive scrutiny. I said two adjacent chain segments drained by two evictors would mean one zeroes the other's tag via `update_links`. On a closer read every chain splice — `condemn`'s neighbour patch, `unlink`, `link_dest_at_head` — runs under the owning `TtlBucket`'s `chain_lock`, held across the whole finalize, and `condemn` captures its links *inside* that critical section while `unlink` reads them live. So a captured link should never name an already-condemned neighbour, and `update_links` should probably never target an `AwaitingRelease` word at all.

That's a structural read rather than a proof — I'd want a reproduction before asserting it either way, which is exactly what I failed to do the first time. **The ask was still right**: carrying the tag in the type is strictly more robust than preserving bits at call sites, it's free, and it removes a whole category of future mistake. But I stated the reachability as established fact when it wasn't, and you acted on it. Sorry.

One consequence worth recording in a comment while you're in there: `try_release_condemned` is a single-shot CAS with no retry, which is safe *because* the only writers of an `AwaitingRelease` word are the three claimants and all of them change the state, so a lost CAS means someone else performed the free. If an `update_links` against a condemned word ever did become reachable, that same lost CAS would become a permanent strand — and nothing sweeps `AwaitingRelease`.

## Rebase: clean, and #68 helps you

`git cherry-pick 5568ad9` onto `main` (`2788fcf`) applies with zero conflicts and no silent compile break. On the interaction I flagged: #68 split `get_pinned` into a converging revalidation retry and an unbounded pin-failure retry, and your change feeds `AwaitingRelease` into the latter. **The termination argument holds and is stronger for your case than for the `Draining` case that loop was written for** — `condemn` runs only after `Segment::clear` has emptied the hashtable of that segment, so a fresh lookup can't resolve back into an `AwaitingRelease` segment, and it converges in at most one re-lookup where `Draining` waits out a whole drain. Worth a sentence in the commit message. The cost is strictly more pin failures, each a snooze plus re-lookup, bounded by the condemn window.

Full gate green on the cherry-pick: workspace tests, `--features debug`, loom (31 models), both fault-injection targets, clippy `-D warnings`, and `--no-default-features` since the new metrics are cfg-gated.

## What we owe you

I mutation-tested your branch. Your `metadata_roundtrip` case is genuinely non-vacuous — dropping the tag in `pack()` fails it cleanly. But three of the four fixes here survive complete deletion with a fully green suite: reverting `update_links` to write `tag: 0` (the original defect), deleting the generation stamp from `cas_condemn` (defect 2 unfixed), and returning the backout to a plain `fetch_sub(Release)` (defect 3 unfixed) all pass, the middle one including all 31 loom models.

That's ours to fix, not yours — we offered the reproductions and they're the coverage for exactly these. Working on it now; I'll push them to the same branch. Don't hold your rebase for it.

brayniac:
The coverage we owed you is ready. **All four mutations now die**, each verified by applying the mutation, capturing the failure, and reverting.

Branch **[`brayniac/cache-rs:repro-for-63`](https://github.com/brayniac/cache-rs/tree/repro-for-63)** now has two commits on top of your `5568ad9`:

```
git fetch brayniac repro-for-63
git cherry-pick 0db454c 33b53d1
```

| Fix | Test | Mutation it kills |
|---|---|---|
| defect 1 | `condemned_segment_is_freed_under_a_live_reader_pin` | `AwaitingRelease` readable again — now fails behaviourally, not just via the static table |
| tag preserved | `update_links_preserves_the_condemn_tag` | `update_links` writing `tag: 0` |
| defect 2 | `a_stale_release_token_cannot_free_a_later_incarnation` | dropping the generation stamp in `cas_condemn` |
| defect 3 | `acquire_backout_must_not_strand_a_condemned_segment` | backout reverting to plain `fetch_sub(Release)` |

Two of the three needed **no seam at all**. The stale-token test drives a real two-incarnation lifecycle through production APIs — reserve, drain, condemn, release, reserve again, drain, pin, condemn — captures the condemned word at the point where a stalled thread's load would happen, and then issues exactly `try_release_condemned`'s compare-exchange against that stale token. It asserts the CAS loses, the segment stays `AwaitingRelease` with its pin intact, and the *real* last reader still completes the handoff afterwards.

**The backout test did need a seam, and I want to flag how it's built because it touches your function.** The increment/re-check window can only be entered when the state changes underneath the acquire, which no single-threaded test can arrange and no multi-threaded one can arrange deterministically. Rather than parameterising `try_acquire_reader` (which would have relocated your function body), it's a `#[cfg(all(test, not(feature = "loom")))]` thread-local fired from two cfg-gated statements inside it. Consequence: **`git diff 5568ad9 -- header.rs` has zero deleted lines** — your signature and body are unchanged, and the hook compiles away entirely outside test builds. It also drives the real `Segments::acquire_item_at`, so the `ReleaseCondemned` arm's `return_segment` and its new metrics are covered too, not only the state transition.

I also fixed two defects in **our own** earlier commit that would have broken your CI: an unused `use crate::sync::Ordering;` and a rustfmt diff at `segments.rs:2427`. Those were ours, not yours. Your `state.rs:177` fmt diff is untouched and is still the only thing `fmt --check` reports.

Gate on the branch: workspace tests, `--features debug`, loom (22), both fault-injection targets, clippy `-D warnings` — and separately `cargo clippy --all-targets` *without* `--all-features`, since `--all-features` enables `loom` and cfg's the new tests out entirely. Worth knowing if you lint locally: the all-features run alone doesn't check this code.

Still yours, unchanged from my last comment: the `cargo fmt` one-liner, and the stale `"remains readable for in-flight pins"` clause in `Segments::condemn`'s doc at `segments.rs:921`. Take the tests or leave them — if you'd rather shape them differently, they're a starting point rather than a requirement.

mux:
Thanks Brian. No need to apologise for the reachability call — I made the same mistake reading your first note. I checked that `update_links` rebuilds the word through `pack()` and that `condemn` splices neighbours after its own CAS, but I never looked at the `chain_lock`, which is the part that actually decides it. The ask was right either way.

I have taken your tests as they are. I also re-applied all four mutations here rather than trusting the table, and each one dies: `is_readable` regaining `AwaitingRelease`, `update_links` writing `tag: 0`, `cas_condemn` losing the generation stamp, and the backout going back to a plain `fetch_sub(Release)`.

The two items you still list as mine are addressed. `pack()` is on one line and `fmt --check` is clean, and `condemn`'s doc no longer says the state remains readable for in-flight pins. I have also added the single-shot CAS note on `try_release_condemned`, and a paragraph on the #68 interaction in the commit message, both as you suggested.

Your clippy point was a real gap on my side. Every clippy run I had done was `--all-features`, so none of the new test code had ever been linted here. Both invocations are clean now.

One for you in return: `pin_failure_tests::merge_churn_no_false_miss_no_resurrection` is flaky under load. 3 failures in 15 full-suite runs on my branch, and 1 in 15 on plain 2788fcf, so it looks pre-existing rather than mine. It only fails in the full run — 12 runs of it in isolation were all clean, which is how I first mis-attributed it to my own change.

Rebased onto 2788fcf and pushed. Full gate green here: fmt, clippy both ways, workspace tests, `--features debug`, loom, and both fault-injection targets.

brayniac:
Merged as `0d053e3`. Thank you — this was a genuinely valuable contribution and I want to be specific about why, since "thanks for the patch" undersells it.

You found a use-after-free by **reading** the code, from a port to a different language, and reported it with the reachability limits and your own failed refutations attached. Every defect held up under independent verification. Two reproduced deterministically; the third turned out to be *more* reachable than you claimed rather than less, because it becomes materially easier to hit once your own first fix lands — which is why the parts had to ship together, exactly as your C port had already concluded.

You also caught two things on our side: that `cargo clippy --all-features` never lints the test code (because `--all-features` enables `loom` and cfgs it out), and a flake in `pin_failure_tests::merge_churn_no_false_miss_no_resurrection` that you correctly attributed to `main` rather than to your branch. **We're treating that second one seriously** — it's our test, it asserts the no-false-miss property, and a previous CI failure in that same test is what led us to a real bug rather than a flaky assertion. It's under investigation now; the answer may be a third false-miss path, in which case you've found us a fourth defect indirectly. Either way it's tracked separately and was never a reason to hold your PR.

For the record, since you may want it for the C port: the outstanding items we deliberately did *not* fix here are the recycle-arm spurious `debug_assert` (which reaches `finalize_drained` as well as `TtlBucket::expire`, and is a nuisance rather than a hazard since the backout's decrement balances the same reader's own increment), and the `try_release_condemned` single-shot CAS dependency you documented — safe today only because the sole writers of an `AwaitingRelease` word all change the state.

If you hit anything else porting this, please do open an issue the same way. A report that says "here is the interleaving, here is what I could not refute, and here is what I may have misread" is worth considerably more than a patch without one.

brayniac:
Thanks @mux - appreciate the PR and contribution!

mux:
I'm glad this helped! I kept thinking that I was missing something in the Rust code that would make these issues only happen in my port. :-)

Issue #65, opened 2026-08-19
segcache: get_pinned's bounded re-validation retry returns a false absent for a live key (second false-absent path)


brayniac:
`get_pinned`'s post-pin re-validation gives up after `RESERVE_RETRIES` (3) mismatches and returns `None` for a key that is demonstrably still live. This is a **second false-absent path**, distinct from the stale-location ABA fixed in #60, and it surfaces as an intermittent CI failure in `numeric_concurrency_tests::cas_incr_stress_exact_accounting`.

Filed as a follow-up from #60. Not fixed there — #60 fixes the hashtable returning a wrong answer; this is the layer above giving up after a *correct* lookup.

## 1. Root cause: confirmed, not inferred

I instrumented every `None` exit in `get_pinned` with a counter and printed them at the test's panic site. Four independent failures, all identical in shape:

```
counter must stay resident (no eviction pressure)
  [lookup_none=0 segid_none=0 ttl_none=0 attempts_none=1 pinfail=0 reval_mismatch=49  calls=7508]
  [lookup_none=0 segid_none=0 ttl_none=0 attempts_none=1 pinfail=0 reval_mismatch=97  calls=20576]
  [lookup_none=0 segid_none=0 ttl_none=0 attempts_none=1 pinfail=0 reval_mismatch=9   calls=11445]
  [lookup_none=0 segid_none=0 ttl_none=0 attempts_none=1 pinfail=0 reval_mismatch=90  calls=9608]
```

Reading that off:

- `lookup_none=0` — the hashtable lookup **never** returned `None`. The key resolves correctly every single time; the hashtable is not involved, and #60's fix is not implicated.
- `segid_none=0`, `ttl_none=0`, `pinfail=0` — not a bad location, not lazy expiry, not a drain window.
- `attempts_none=1` — exactly one `attempts >= RESERVE_RETRIES` giveup, and it is the one that produced the panic.
- `reval_mismatch` in the tens — the post-pin re-validation loses the race routinely; the failure is the tail where it loses **three times in a row** in a single `get_pinned` call.

So: the lookup succeeds, the pin succeeds, and `get_pinned` throws the result away because the key moved three times while it was working. `segcache.rs`:

```rust
drop(guard);
attempts += 1;
if attempts >= RESERVE_RETRIES {
    return None;          // <-- false absent: the key is live and resolvable
}
```

[...]

Issue #61, opened 2026-08-19
CI: ThreadSanitizer job for the concurrent paths


brayniac:
## Context

One hardening line produced seven critical concurrency fixes. Most were found by adversarial review, not by tests. Verification of the fixes has leaned on watchdogged stress loops that reproduce rare interleavings by scheduler luck — the most recent bug (unguarded ABA in read-path key verification) reproduces at roughly **1 in 2,400 runs at 24-way parallelism, ~1.3e-9 per lookup**. That is bad economics: slow to run, and a green run is weak evidence.

That bug is formally a **data race**: `SegmentsVerifier::verify` reads item bytes out of mmap'd segment memory non-atomically while another thread rewrites those bytes in a recycled segment. ThreadSanitizer flags racing *accesses* regardless of whether the rare window produces a visible wrong answer, so it does not need the interleaving to line up. This issue reports an actual TSan spike against `main`.

## 1. Does TSan build and run this workspace? Yes.

Exact working invocation (host `aarch64-apple-darwin`, nightly `1.97.0-nightly (d7f14d3d8 2026-05-15)`):

```bash
rustup toolchain install nightly
rustup component add rust-src --toolchain nightly   # required for -Zbuild-std

RUSTFLAGS="-Zsanitizer=thread" \
  cargo +nightly test -p segcache \
    -Zbuild-std --target aarch64-apple-darwin --features debug
```

Notes:
- `-Zbuild-std` needs the `rust-src` component; without it the build fails immediately. It was not installed on my machine.
- `--target <host-triple>` is mandatory — `-Zbuild-std` is a no-op without an explicit `--target`, and the sanitizer runtime will not be linked.
- Cold build including a sanitized `std`: **42s**. No source changes required.
- **No crate-level blockers.** `memmap2`, `crossbeam-utils`, `crossbeam-deque`, `ahash`, `metriken` all built and ran clean under `build-std` + TSan. This was the thing I most expected to break, and it did not.
- `loom` and TSan are mutually exclusive within a single build, but they are separate cargo invocations, so this is a non-issue in practice.
- Supported TSan targets are `{x86_64,aarch64}-{apple-darwin,unknown-linux-gnu}`. A CI job would be Linux-only, matching the existing expectation.

## 2. Signal-to-noise: excellent. Zero false positives.

Full `--features debug` suite on `main` (e3d83c3): **7 reports, all one race class, 0 false positives.**

```
WARNING: ThreadSanitizer: data race

[...]

PR #66, opened 2026-08-19
Bump segcache version to 0.4.3

brayniac:
segcache-only bump: **0.4.2 → 0.4.3**. `keyvalue` stays at 0.3.1 — `crates/keyvalue/` is byte-identical to the tree published as 0.3.1 (`git diff e3d83c3..main -- crates/keyvalue/` is empty), so bumping it would republish the same code under a new number.

Rebased onto current `main` immediately before merge, and `crates/segcache/Cargo.toml` was confirmed to read `0.4.2` on `main` at that moment. (A release branch cut before another PR that touches the same manifest can silently *revert* the version through a squash merge, and nothing in CI catches it.)

## What's in this release

**Correctness — memory safety**

- **#63 — reader-pin reclamation of condemned segments.** A use-after-free: a segment condemned while readers still held pins could be reclaimed out from under them. Reported and fixed by external contributor **@mux**. This is the most serious fix in the release.
- **#60 — read-path stale-location guard (ABA).** `SegmentsVerifier::verify` compared key bytes by reading raw segment memory at the location loaded from the bucket slot, with no pin and no generation tag. A reader that stalled while the item relocated and its segment was recycled and rewritten verified against the *new* occupant, concluded "different key", and returned `None` — a false-absent, returned before any retry logic could see it. All six verify sites now route through one helper carrying a named invariant, enforced as a debug-build precondition.
- **#68 — `get_pinned` revalidation convergence.** The bounded revalidation retry could give up and return `None` for a live key under heavy same-key churn (issue #65). Retries now converge on the key's new location instead of re-racing the stale one.

**Correctness — accounting**

- **#59 — relocation gauge accounting.** Both relocation paths (`s3fifo_promote_from`, and `copy_into`'s early return on `RelinkFailure`) re-homed item header counts without re-adding the item gauges, permanently under-counting. Restructured to per-item compensation so no pending state remains for a future early return to drop.
- **#59 — directional `check_integrity`.** The exact item-count match asserted an invariant the design does not hold (unpinned unlinks deliberately skip decrements), making it a latent `--features debug` flake. Relaxed to the honest directional form.
- **#71 — `item_dead` / `item_dead_bytes` are now true gauges.** They had no decrement site, so they were cumulative counters wearing gauge labels. They now measure current reclaimable dead space and come back down when a segment is reclaimed.
- **#75 — `clear_time` / `expire_time` / `evict_time` are timed on a real clock.** All three measured sub-millisecond operations with the *coarse* (1-second) `Instant`, so `elapsed()` returned either `0` or exactly `1_000_000_000` ns. Not dead counters — randomly valued ones. Fixed in #77.

**Testing and tooling**

- **#67 — loom coverage of the hashtable slot protocol** via a stateful key oracle.
- **#70 — `get_hit` benchmarks**, and `get` renamed to `get_miss` so the two paths are distinguishable.
- **#59 — new `fault-injection` feature.** Non-default, deliberately *not* implied by `debug` (`debug` is observational; this changes behaviour), `#[doc(hidden)]`, and covered by its own scoped CI steps.

## Metrics behaviour — read this before upgrading

Operators scraping these will see **step changes**. All four are corrections, not regressions, but they are visible.

1. **`item_current` / `item_current_bytes`: a second step change, in the *opposite* direction to 0.4.2.** 0.4.2 corrected an **upward** drift (unpinned unlinks skipping decrements with nothing reclaiming the residual). This release corrects a **downward** drift under S3-FIFO (#59) — relocation re-homing counts without re-adding the gauges. So a deployment that upgraded through both sees the gauge step *down* at 0.4.2 and *up* at 0.4.3.
2. **`item_dead` / `item_dead_bytes` change meaning** (#71). Previously monotonically increasing (a counter in gauge clothing); now a true occupancy gauge that rises and falls with reclaimable dead space. Any dashboard treating them as ever-increasing — a `rate()`, a `counter` panel — needs changing.
3. **`clear_time` / `expire_time` / `evict_time` start reporting real numbers** (#75). Previously `0` or a whole spurious second per tick-straddling operation; `evict_time / segment_evict` looked like a latency and was a random walk. Historical values for these three are garbage and should not be compared against post-upgrade values.

## Residuals — scheduled work, with tracking issues

Named honestly, but none of these are unowned gaps:

- **The ABA class in #50 is not closed by this release.** #60 makes the *observable outcome* correct (detect and retry); it does not eliminate the racing access. The route to eliminating it — generation-tagged `Location`s — is **designed and deliberately held out of this release** for a measured performance reason, and tracked in **#50**. The fix is written and deferred, not missing.
- **ThreadSanitizer still reports races on the concurrent paths**, for exactly that reason: detect-and-retry leaves the access in place. Making a TSan CI gate viable depends on the same #50 work and is tracked in **#61**.
- **`add` / `replace` remain check-then-act.** Making them atomic needs conditional-insert primitives in the engine, which do not exist yet; until then the check and the act are separable under concurrency.
- **#76 is an open flake report we could not reproduce.** `pin_failure_tests::merge_churn_no_false_miss_no_resurrection` was reported flaky; 410 runs across a 2×2 of profile × integrity did not reproduce it. Left open pending the panic text rather than closed as unreproducible.

https://claude.ai/code/session_01MDkBFNHhYhbpGBgpGDQwmw


brayniac:
**Published, and the affected earlier releases are now yanked.**

`segcache` **0.4.0**, **0.4.1**, **0.4.2** and `keyvalue` **0.3.0** are yanked. All three segcache 0.4.x carry the reader-pin reclamation use-after-free fixed in #63; 0.4.1 and 0.4.2 additionally carry both confirmed false-absent bugs (#60's stale-location verify ABA and #68's bounded-revalidation giveup), and 0.4.0 predates the lazy-expiry checks as well. `keyvalue` 0.3.0 carries the non-atomic numeric read-modify-write fixed in #48 — 47,336 of 80,000 concurrent increments survived in the reproduction.

Yanking stops new resolution selecting them; anything already pinned in a `Cargo.lock` continues to resolve and build unaffected.

**Deliberately not yanked: `segcache` ≤ 0.3.0 and `keyvalue` ≤ 0.2.0.** Those are the pre-concurrency `&mut self` API, and every defect fixed in the 0.4.x line is a concurrency bug reachable only through `&self` sharing — which those versions' types make impossible. They are sound for single-threaded use and yanking them would penalise users of a working release.

**0.4.3 is the only supported 0.4.x.** Residual engine work is tracked and deliberately not in this release: #50 (ABA elimination, designed and held for a measured hot-path cost — also the route to a green ThreadSanitizer run per #61), #72 (whether the location tag can subsume the post-pin revalidation), and #76 (an unreproduced flake report, 0 failures in 410 runs across a 2×2 of profile × integrity, awaiting the reporter's panic text).

PR #68, opened 2026-08-19
segcache: converge get_pinned's revalidation retries on the key's new location

brayniac:
## Summary

`get_pinned`'s post-pin revalidation gave up after `RESERVE_RETRIES` (3) mismatches and
returned `None` for a key that was demonstrably still live (#65). The retry budget was
being spent **re-racing the lookup from scratch**: on a mismatch the revalidation lookup
has already returned the key's *new* location, and the old loop threw it away and looked
the key up again.

This makes the retries **converge** on that returned location, and gives the bound its
own, larger constant now that each attempt is both cheaper and less likely to lose.

## Mechanism, confirmed against the code

The old loop, per attempt: `lookup(key)` → `acquire_item_at` (pin) → `lookup(key)` again
→ compare. A mismatch means another thread published a new location for this key inside
that window, so:

- the **vulnerable window** spanned *lookup + pin + lookup*;
- the mismatch also *discarded* the one piece of currently-true information it had just
  learned — where the key actually is;
- and three of those in one call returned `None` for a live key.

Following that returned location instead (`follow_republished`) shrinks the per-attempt
window to *pin + lookup* and removes the second lookup per retry entirely. That second lookup is the cost
flagged as "optimize later" when item 7f's revalidation landed — confirmed against the
code: `get_pinned` performed two `lookup`s on **every** call, and the fast path still
performs those two on a hit, with retries now costing one each instead of two.

## The fix against #65's four constraints

1. **The revalidation still exists, unweakened.** Every location handed out is still one
   the reader pinned *and then re-confirmed by a fresh hashtable lookup*. The
   aliased/torn read 7f defends against is unreachable by exactly the same argument as
   before.
2. **The re-lookup-from-scratch is still the soundness argument.** The location
   `follow_republished` returns is one *the fresh revalidation lookup itself just
   returned*, so it is currently-published by the same "stale entries are cleared before
   recycle" clause. Following it is that argument applied, not a substitute for it.
3. **Neither rejected shape is reintroduced.** The pinned location is never trusted
   without a lookup — after following a republication, the very next thing that happens
   to it is another pin *and another full revalidation lookup*. No generation recheck.
4. **The bound is kept.** `REVALIDATE_RETRIES` is a hard cap, and
   `bounded_giveup_when_every_revalidation_loses` fails (as a watchdog wedge) against any
   build that removes it.

### Why 16 and not 3

Exhausting the budget now requires 16 *consecutive* republications of one key, each
landing inside a window that is roughly half its former size. At the ~1%/attempt mismatch
rate #65 measured on the worst realistic workload that is ~1e-32, against ~1e-6 at 3. The
worst case remains bounded: 16 pins and 17 lookups.

### One deliberate behaviour change beyond the retry shape

If the revalidation lookup returns `None` (key not published anywhere), `get_pinned` now
returns `None` immediately instead of re-running the outer lookup up to three times. The
revalidation *is* a fresh from-scratch lookup, so its `None` is exactly the answer the
top-of-loop `?` produced; the get simply linearizes at that lookup. Frequency-bump
behaviour is unchanged (every from-scratch lookup still honours `update_freq`, including
the uncounted pin-failure retries).

### Deterministic — red before, green after

`crates/segcache/src/revalidation_tests.rs`, driven by two new `fault-injection` hooks
that stand in for the racing writer at the two points that matter (`after_lookup` fires
per *from-scratch* lookup; `before_revalidate` fires inside the pin→revalidate window).
No production behaviour outside `feature = "fault-injection"`.

`get_converges_instead_of_re_racing_the_lookup` republishes the key after **every**
from-scratch lookup, forever. The pre-fix loop re-arms that hook on every attempt and
burns its whole budget; the converging loop fires it exactly once.

Against the pre-fix `get_pinned` body (hooks and tests unchanged), all three are red:

```
running 3 tests
test revalidation_tests::budget_absorbs_republication_inside_the_revalidation_window ... FAILED
test revalidation_tests::bounded_giveup_when_every_revalidation_loses ... FAILED
test revalidation_tests::get_converges_instead_of_re_racing_the_lookup ... FAILED

get_converges_instead_of_re_racing_the_lookup:
  false absent: the key was republished, never removed, and every lookup resolved it
  — a bounded retry that re-races from scratch turns that into a miss (#65)
budget_absorbs_republication_inside_the_revalidation_window:
  assertion `left == right` failed: the hook must have spent the whole budget minus one
    left: 3   right: 15
bounded_giveup_when_every_revalidation_loses:
  assertion `left == right` failed: give-up must cost exactly the budget
    left: 3   right: 16

test result: FAILED. 0 passed; 3 failed
```

After the fix:

```
running 3 tests
test revalidation_tests::get_converges_instead_of_re_racing_the_lookup ... ok
test revalidation_tests::bounded_giveup_when_every_revalidation_loses ... ok
test revalidation_tests::budget_absorbs_republication_inside_the_revalidation_window ... ok

test result: ok. 3 passed; 0 failed
```

### loom

`loom_revalidation_retry_survives_republication` (`hashtable/table.rs`), using a new
`ChurnOracle` — the `RecyclingOracle` idea extended from one relocation to a chain of
republications, sequenced in production order (bytes valid → relink → source recycled).

**Scope, stated plainly:** the reader is a *transcription* of `get_pinned`'s retry loop,
not a call into it — the pin reads `Segments`' mmap'd headers, which are not loom types.
What is real: the hashtable operations, their interleavings against the republications,
and `REVALIDATE_RETRIES` itself, which is imported from the production module.

It asserts two things: no false absent, and `lookups <= CHAIN.len() + 1` — one
from-scratch lookup plus one revalidation per attempt, which *is* the convergence
property.

**Proven to fail, twice** (not a control test):

- neutering `REVALIDATE_RETRIES` to 2 (production constant): `false absent: the key was republished 2 times and never removed, so every lookup
  could resolve it...` — found in 0.02s
- restoring the pre-#65 from-scratch reader: `each retry must follow the location the revalidation already returned: 6 lookups
  for 2 republications means an attempt re-raced from scratch` — found in 0.02s

Full loom suite: 25 models, 52.3s wall (the new model is 37.6s standalone, and runs concurrently with the rest).

### Benchmark A/B — no regression detected, bounded by an A/A control

**Method.** Both arms built once into separate target dirs (no compile between
measurements), run **interleaved**, on a **quiet machine** (1-min load average 6-12 on 32
cores, recorded at every single run). Statistic is the **paired per-round ratio**
(branch/main within the same round), which cancels slow drift. Every claim below is
calibrated against an **A/A control**: the same `main` binary on both sides, 3 interleaved
rounds, where the true effect is zero by construction — so its spread *is* the noise floor.

| benchmark | A/A control (same binary) | noise band | main (ns) | branch (ns) | effect per round | median |
|---|---|---|---|---|---|---|
| `get/1b/0b` | +0.87, −0.78, +1.26% | ±1.26% | 16.93 16.34 16.44 16.66 | 17.10 17.33 16.56 16.55 | +1.04, +6.03, +0.73, −0.68% | **+0.88%** |
| `get/255b/0b` | −0.15, −1.14, +2.87% | ±2.87% | 32.41 31.64 32.37 32.27 | 33.06 31.50 31.86 31.52 | +1.98, −0.43, −1.57, −2.30% | **−1.00%** |
| `get_hit/1b/0b` | −1.59, +0.19, +1.77% | ±1.77% | 40.73 38.80 38.61 40.23 | 40.86 39.81 39.89 39.21 | +0.30, +2.62, +3.32, −2.53% | **+1.46%** |
| `get_hit/255b/0b` | −1.49, +0.87, +1.51% | ±1.51% | 83.71 79.34 81.07 82.14 | 82.63 81.40 82.52 81.78 | −1.29, +2.60, +1.79, −0.44% | **+0.68%** |

**Every median effect falls inside that benchmark's own same-binary control band, on all
four.** So: no regression detected, and the measurement bounds any real regression at
roughly the control width — about ±1.3% on `get/1b/0b`, ±1.8% on the 1b hit path. I am
**not** claiming the change is exactly free: a sub-1% effect is still below what this
setup resolves. Individual rounds disagree in sign (`get/1b/0b` shows +6.0% in one round
and −0.7% in another) which is what a null result looks like at this resolution.

**Machine load is the whole story for this benchmark.** An earlier attempt ran while the
box sat at load 83-101, where the *same-binary* A/A control produced apparent per-round
effects from −13.0% to +2.6% — a noise floor ~10x wider than the 1-2% that matters here
(#60's first cut cost exactly that on this function). Those numbers were discarded as
unmeasurable, and the table above is the re-run on a quiet box, where criterion's own
confidence intervals tightened from ±3% to ±0.15%. Anyone repeating this should check
load first; the two runs differ by 3x in absolute ns.

An intermediate, worse-controlled run suggested a consistent ~+3% on `get_hit/1b/0b`, and
I restructured the code in response. The control shows that apparent effect was inside
the noise. **The restructure is not credited with fixing a regression** — it stands on its
own as the simpler shape: one loop, direct `return` on the hit, both retry decisions
`#[cold] #[inline(never)]` on edges the hot path does not take. That is #60's lesson
applied by construction rather than discovered after the fact.

#### Note on what these benchmarks cover

`get/1b/0b` and `get/255b/0b` **never insert anything**, so they only exercise the MISS
path (`lookup(key)?` → `None`) and never reach the pin or the revalidation this change
touches. `get_hit/*` is an ad-hoc group (1024 resident keys, **not committed**) added
identically to both arms to cover the hit path at all. A real hit-path benchmark is worth
landing separately — its absence is why the repo's `get` benchmarks could not have caught
a regression in this function either.

### Stress — corroboration only

The reproduction the issue documents, run interleaved on one 32-core machine, 24-way
oversubscribed, both arms alternating so each sees identical load. Every run wrapped in
`timeout 120`, so a hang would count as a failure:

| round | `main` (c541316) | branch |
|---|---|---|
| 1 | 7 / 350 | 0 / 350 |
| 2 | 12 / 350 | 0 / 350 |
| 3 | 11 / 350 | 0 / 350 |
| 4 | 8 / 350 | 0 / 350 |
| 5 | 14 / 350 | 0 / 350 |
| **total** | **52 / 1750 (3.0%)** | **0 / 1750** |

Zero timeouts on either arm — the branch's 1750 clean runs are also 1750 pieces of
evidence that the bounded retry does not turn into a spin.

Every `main` failure is `numeric_concurrency_tests.rs:113`, "counter must stay resident
(no eviction pressure)" — the false absent, not some other assertion.

This is corroboration, not the primary claim: 1750 runs cannot distinguish "fixed" from
"made 100x rarer". The deterministic tests and the loom model are what make the
structural claim.

### CI (this branch, run 32290125027)

| job | result |
|---|---|
| `build-and-test (ubuntu-latest)` | pass |
| `build-and-test (windows-latest)` | pass |
| `build-and-test (macos-latest)` | pass |
| `lint` | pass |

macOS matters specifically: it is the runner `cas_incr_stress_exact_accounting` has been
failing on (twice on #60, twice on #67), and the runner whose timing hid the original 7f
bug that Ubuntu and Windows exposed. Green on all three is the standard #65's comment
asks for.

Local gate, all green: `cargo test --workspace`; `-p segcache --features debug`;
`--features loom -- loom` (25 models); `--features fault-injection --lib` (133 tests) and
`--features fault-injection,debug --test copy_into_relink_failure`;
`clippy --all-targets --all-features -- -D warnings`; `fmt --check`.

## Not changed

`cas` and `try_into_numeric` have the same bounded revalidation, but they give up with
`Exists`, never a false `NotFound`, and they revalidate through `lookup_slot` (they need
the slot as well as the location), so converging them is a separate change with no
correctness pressure behind it. Left as-is.


brayniac:
**Correcting the precision claim in this PR's benchmark section.** The conclusion stands; the stated resolution was optimistic.

This PR reported an A/A control of **±1.26%** on `get_hit/1b/0b`, four effect medians inside their control bands, and any real regression "bounded at roughly ±1.3-1.8%".

An attribution study on #78 has since shown that **the A/A control cannot observe code-layout noise**, because both arms of an A/A comparison share a layout. Measured directly: six padding-only variants of a single binary — semantically identical, hot-path instructions identical, differing only by `global_asm!(".space N")` — span **0.6-1.3 ns (1.6-3.4%)** on this same benchmark, on a quiet machine where the A/A control read 0.097 ns (0.25%). So an A/A figure can advertise a resolution roughly an order of magnitude better than a single-binary comparison can actually achieve.

**What that does and does not change here:**

- **The conclusion is unaffected, and if anything more robust.** "No regression detected; the effect is below resolution" becomes *more* true when the true floor is wider than the floor we measured against.
- **The bound was too tight.** "Bounded at roughly ±1.3-1.8%" understates the real uncertainty for a single-binary comparison at this scale. Anyone citing that number as a precision claim should use the layout spread instead.

**Corrected protocol for effects near ~1 ns**, which we are adopting: build each variant as several padding-only layouts and compare per-layout minima, rather than trusting a single-binary A/A control. Where the question permits, attribute from **disassembly instead of timing** — that settles instruction-form questions exactly, with no benchmark involved.

Flagging it here rather than leaving the number to be quoted later. It is the same failure mode this line of work kept turning up: a real measurement that could not observe the thing it was being cited for.

Issue #72, opened 2026-08-19
segcache: can the location generation tag subsume get_pinned's post-pin revalidation lookup?


brayniac:
Speculative optimisation with a real payoff and a real gate. Filed separately rather than bundled into #50, because it trades an exact check for a probabilistic one on the read hot path, and that deserves to be evaluated on its own merits rather than riding a correctness PR.

## The idea

`get_pinned` currently pays a **full second lookup** on every hit: after pinning, it re-looks-up the key and confirms it still resolves to the same location (the read-path guard from #60, refined by #68's converging retries). Once locations carry a generation tag (#50), the tag proves the segment has not been recycled since the location was published — which is the memory-safety property the revalidation was written to establish.

If the tag suffices, the revalidation lookup can go, replacing a lookup with a compare. #50 currently costs ~4% on `get_hit/1b/0b`; removing a full lookup could put the read path **ahead** of today's baseline rather than behind it.

## Why this is NOT the generation recheck that was already tried and rejected

Important, because the 7f history makes this look like a re-proposal of a known-bad fix. It isn't, and the difference is the whole argument.

The earlier attempt captured the segment generation **post-lookup** and re-checked it **post-pin**. That left a residual under heavy preemption: a recycle could land between the lookup's key verification and the generation *snapshot*. Two independent reads, therefore a TOCTOU between them. The fresh-lookup revalidation is what replaced it.

The tag has no such gap: **it rides inside the location, published atomically with the entry**. There is no separate sample that can be stale — the tag is read out of the same slot word the lookup already loaded, and compared against the header generation read *under the pin*, which itself prevents recycle. So "tag matches" genuinely establishes "this segment has not been recycled since this location was published", closing the hole the earlier attempt left open.

## What it does NOT subsume — this is the gate

**The revalidation also establishes "still published"; the tag says nothing about it.** A key deleted or relocated between the lookup and the pin leaves the segment un-recycled, so the tag matches and the reader proceeds.

- **Relocation** is harmless: the bytes are unchanged, so a stale-but-valid location returns the correct value.
- **Deletion** is the question. The reader would return a value for a key whose hashtable entry was already unlinked. That is arguably a legal linearization — the `get` began before the `delete` — but it is precisely the acked-delete-resurrection *shape* that #55 was written to eliminate, and `add`/`replace` are built on presence probes (`get_no_freq_incr().is_some()`), so the user-visible semantics need deriving rather than assuming.

[...]

PR #78, opened 2026-08-20
segcache: generation-tagged locations (fixes #50)

### LINKED ISSUE #50: segcache: raced-old/rollback unpinned unlink can pin a recycled segment incarnation (Location has no generation)

### LINKED ISSUE #69: segcache: s3fifo_promote_from has no regression coverage — breaking it entirely leaves the suite green
Found by bite-check while building #50, and true on `main` today, independent of that work. Filing separately so the gap is recorded as having existed rather than only implicitly closed by the test #50 adds.

## The gap

Deliberately breaking the relocation in `Segment::copy_into` reddens three tests (`merge_evict_copies_survivors_into_spare`, `merge_compact_combines_under_full_segments_into_spare`, `merge_halts_at_pinned_candidate_and_relocates_survivors`).

Deliberately breaking the equivalent relocation in `Segments::s3fifo_promote_from` — the same class of change, in the sibling function — leaves **the entire 134-test suite green**.

So S3-FIFO promotion can degrade to a complete no-op and nothing anywhere notices. Items would simply stop being promoted from the small pool into main; the cache keeps working, hit rates degrade, and no test, assert or metric fails.

## Why this is worth its own issue

This is the **second** silent-degradation defect found in this exact function this week. The first was the missing `ITEM_CURRENT`/`ITEM_CURRENT_BYTES` compensation fixed in #59 — where the promote path re-homed an item's header count without re-adding to the global gauges, unlike `copy_into`'s batch compensation, producing a permanent downward drift.

Two independent defects in one function, both silent, both found by inspection rather than by a failing test, is a pattern rather than a coincidence — and the common cause is that nothing exercises the function's *effect*. #50 adds a deterministic test asserting promotion actually relocates items, which closes it going forward, but the gap predates that change and is worth tracking on its own.

## The method that found it, which generalizes

The audit is cheap: break a production function deliberately, run the suite, and see what goes red. If nothing does, that function's behaviour is unverified no matter how many tests appear to touch the path around it. `copy_into` vs `s3fifo_promote_from` differ only in whether anything asserts the outcome, and the comparison is what made the gap visible.

Worth running across other production paths — eviction policies, the expire/flush paths, and the merge/compact decision logic are the obvious candidates, since all three are "correct behaviour is a policy outcome" rather than "wrong behaviour crashes". Related in spirit to #67's per-model neutering discipline, applied per-function instead.

brayniac:
> **⚠️ Performance figures below are superseded.** The "+1.2 ns / +2.9%" result was measured against an A/A control that **cannot observe code-layout noise** (both arms share a layout). Padding-only variants of one binary span 0.6–1.3 ns on this benchmark, so that control advertised a floor roughly 10× tighter than achievable. Corrected attribution: the tag check costs **+0.94 ± 0.26 ns**, but this change **net versus its parent is +0.28 ± 0.24 ns — indistinguishable from zero**, because the `Location`-taking signature lets LLVM hoist the item-pointer computation out of the bucket-scan region (per-iteration register reloads 7 → 2). See [the correction comment](https://github.com/pelikan-io/cache-rs/pull/78#issuecomment-5357283176).

Fixes #50. Closes #69. **Targets 0.4.4, not 0.4.3** — this hardens against a class that is currently theoretical and carries a measured read-path cost, while 0.4.3 fixes bugs that are confirmed and live.

## The problem

A `Location` is `(segment id, offset)` with no incarnation identity. A segment can be drained, recycled, re-reserved and refilled while a thread holds a location naming it, so a stale location can silently resolve to a *different item at the same address*. That underlies the unpinned-unlink accounting hazards, the `DifferentKey` and stale-`SlotRef` residuals, and it is why a TSan gate is currently impossible (#61) — `verify` reads segment bytes whose validity is only established after the read.

## The change

`Location` is repartitioned inside its existing 44 bits:

```
before:  [ 24-bit segment id ][ 20-bit offset>>3 ]
after:   [ 18-bit segment id ][ 6-bit tag ][ 20-bit offset>>3 ]
```

The tag is the low 6 bits of the segment header's `generation`. Because it rides **inside** the location, which rides inside the packed hashtable slot word, every existing compare-exchange on a published entry validates it with no new code and no extra load — a stale-incarnation CAS simply fails, and each site's existing failure arm already handles that. That is the main argument for tagging the location rather than carrying identity beside it.

`Segments::resolve(location)` compares the tag against the live header generation for the sites that need an explicit check; `None` always means "this location is no longer yours", never an error.

### ⚠️ Breaking: capacity

Maximum addressable segments drops from 16,777,215 to **262,142** — 256 GiB of heap at 1 MiB segments, 2 TiB at the 8 MiB maximum. Configurations above that now **fail at construction** with an error naming the limit and pointing at `segment_size` as the lever, rather than silently aliasing. The top id is deliberately unissuable so `Location::GHOST` cannot be aliased by construction.

### Prerequisite: the generation bump moved

`generation` was bumped in `try_reserve` (`Free → Reserved`). But `try_release` returns a `Reserved`/`Linking` segment straight back to `Free` with no fill, seal, drain or recycle, and is reachable from both chain-extension election-loser paths — so a *failed election* advanced the counter for free, at a rate decoupled from segment lifecycles. A 6-bit tag cannot afford that.

The bump now lives on the two transitions that end a *used* incarnation — `try_release_drained` (`Draining → Free`) and `try_release_condemned` (`AwaitingRelease → Free`) — with `try_release` (`Reserved|Linking → Free`) as the visibly-named non-bumping exception. It sits on the **state transition**, not in the queue helper, because `SegmentGuard::drop` frees a condemned segment by pushing to the free queue directly (it holds only a raw pointer, not `&Segments`), so a queue-level bump would have silently skipped a condemned free.

**This bump site now carries two independent invariants**, and a future change moving it back would look locally harmless while breaking both:

1. the tag's collision hardness (this PR), and
2. the bound on `get_pinned`'s stale-incarnation retry arm being expensive to consume — an attempt is charged only when `resolve` fails, which requires a real recycle. Under the old site, a burst of lost elections could have consumed that budget with no lifecycle at all, starving a live key into a false absent.

`tests::generation_advances_on_exactly_the_two_transitions_that_end_a_used_incarnation` pins this as a full transition table.

## Why 6 bits, stated honestly

Aliasing requires one thread to stall across **64 lifecycles of one specific segment id**. The earlier draft claimed this cost 64 segments' worth of writes; that was wrong and is not the argument. A lifecycle can be cheap — `drain_chain` drains the Live tail directly for `clear()`/`expire()`, and a `Sealed` segment is freed as soon as `live_items() == 0`. "Same offset" is also *not* an independent coincidence: segments are append-only from a fixed start, so with uniform item sizes the n-th item of every incarnation lands at the same offset. And at the `remove_at`/`rollback_reservation` sites the tag is the only remaining check — no slot term.

What remains is a real, bounded residual: 64 lifecycles inside one stall window. Tracked, not hidden.

## Performance: measured, with its resolution

Counterbalanced interleaved A/B (`e01a365` vs `22639b7`), 4 rounds per arm with arm order rotated, preceded by a 4-run A/A control on the baseline binary. Criterion defaults; shared machine at load 5.8–10.6.

| group | effect | A/A control | verdict |
|---|---|---|---|
| `get_hit/1b/0b` | **+2.9% (+1.2 ns, 40.7 → 41.9 ns)** | 1.39% | measurably present, 4/4 rounds same direction |
| `get_hit/255b/0b` | +0.39% | 2.15% | not resolved — as a fixed 1.2 ns cost predicts |
| `set/1b/64b` | +0.29% | 3.64% | inconclusive at resolution achieved |
| `incr/hot_counter` | −1.69% | 2.02% | inconclusive at resolution achieved |

The two `get_hit` groups form an **internal consistency check**: a fixed cost must show roughly half the relative effect at 255b (≈2× baseline) that it shows at 1b. It does — a *larger* relative effect at 255b would have proved the sweep was measuring the machine. An earlier sweep failed exactly that test and was discarded rather than published.

Criterion's own verdicts are not usable at this resolution: the A/A control produced a confident **"Performance has regressed. [+2.31%]"** comparing the baseline binary *to itself*.

An earlier version of this branch measured ~4%. Profiling the disassembly showed the tag check routed through a helper that re-derived the header address from `self` while the caller already had it live — a two-deep dependent L1 chain plus a redundant bounds check and panic edge, all stalled behind the `ldar`. Passing the header already in hand collapses it to a single `ldrh`. Semantics are preserved by identity: `self.header(seg_id)` *is* the reference the caller holds.

Honest caveat: +1.2 ns exceeds what one L1 load and compare should cost (~0.3–0.6 ns), and this diff is not only the tag check — `pack_location` now stamps a generation on the publish path, and the field split changed the unpack shifts.

## Coverage, scoped honestly

- **The loom models do not cover the new pin-failure arm.** The existing model transcribes the retry loop *without* the pin, and the tag triage lives on the pin-failure edge, so `loom_revalidation_retry_survives_republication` passing unmodified shows the modelled loop was not disturbed — **not** that the new arm is correct. That arm's coverage is deterministic-only.
- `revalidation_tests::budget_absorbs_recycled_incarnations_without_a_false_absent` drives 15 republish-**and-recycle** cycles inside one `get` and asserts no false absent, plus that the budget was charged a non-zero number of times (15 of 16), so it cannot decay into a no-op. Proven red four ways, including a republish-only variant that charges nothing — the vacuity trap.
- `s3fifo_promotion_relocates_items_into_the_main_pool` closes #69: before it, breaking `s3fifo_promote_from` entirely left the whole suite green.
- `replace_at` now gates `old_location` on the incarnation tag, matching `delete`. A remover pin freezes the generation from the moment it is taken; it does not prove the pinned segment is the incarnation the location names.

## Residuals

- The tag is probabilistic (6 bits), not exact — see above.
- #72 asks whether the tag can subsume `get_pinned`'s post-pin revalidation lookup, which would remove a full lookup from the hit path and could put this ahead of baseline. It has a real gate (the revalidation also establishes "still published", which the tag does not) and is deliberately not bundled here.
- Eliminating the unpinned unlinks entirely is separate work: tagging makes them *safe*, eliminating them fixes the distinct missed-decrement accounting gap.


brayniac:
**Correcting this PR's performance claim: the +2.9% figure over-attributes, and #50's net cost is indistinguishable from zero.**

A follow-up attribution study measured the components. Summary:

- **The tag check itself: +0.94 ± 0.26 ns** — positive in every sweep and every layout, consistent with ~4 cycles of serialized load-use latency. That part of the original number is real.
- **#50 net, versus its parent `7fe80ca`: +0.28 ± 0.24 ns** — i.e. **not distinguishable from zero**, and not the +1.2 ns / +2.9% reported above.

The gap is a win nobody predicted. The `Location`-taking signature lets LLVM hoist the item-pointer computation out of the bucket-scan region instead of keeping `segment_size`, `data`, `seg_id` and `location` all live across the scan loop and the `memcmp`. Per scan iteration the parent reloads seven registers; this branch reloads two, and the success tail shrinks from nine instructions to two. So the rest of #50 pays for most of the tag check.

**Why the original measurement was wrong, and it is not a machine-noise story.** At this effect size, a single-binary A/B measures **code layout** as much as code. Six padding-only variants (`global_asm!(".space N")`, N = 0..80 — semantically and hot-path-instruction-identical) span **0.6–1.3 ns (1.6–3.4%)** on `get_hit/1b/0b`. That is larger than the effect being attributed.

Critically, **an A/A control cannot see this**: the A/A here read 0.097 ns / 0.25%, because both sides share a layout. The machine was quiet; the *layout* was the floor. Our protocol's control therefore reported a resolution roughly 10× better than the one actually achieved.

The corrected protocol for read-path work at this scale is to run each variant as several padding-only layouts and compare the per-layout minima. That is what produced the numbers above.

**Component findings, for the record:**
- The field split (`seg_id(24)|offset` → `seg_id(18)|tag(6)|offset`) is **exactly free** — decidable from instruction forms, not timing. Same two instructions (`lsr`, `ubfiz`), same forms, constants still immediates. The tag sits *between* the other two fields, so extracting either neighbour is still one instruction.
- `pack_location`'s generation stamp is **not on the read path at all**.
- The tag check's codegen is already optimal: `Location::tag()` is not merely inlined but fused — the two 6-bit masks collapse into a single `tst`. Four instructions, no call, header address already in a register.

The check cannot be made cheaper without moving it out from behind the `LDAR` that makes it correct — which is forbidden, since the generation must be read *under* the pin. Two zero-semantic-change cleanups the disassembly suggested (eliminating a duplicated unpack, and a duplicated generation load in `token_for`) were implemented and measured at −0.014 ns and +0.162 ns respectively: nil, because both sit off the critical dependency chain. Neither is proposed.

**No code change follows from this.** The correction is to the claim, not the code — but the 0.4.4 release notes should carry the corrected figure rather than the +2.9% above.

Issue #79, opened 2026-08-20
segcache: builder does not cap segment_size at the location offset field's 8 MiB, so a larger segment silently aliases two items onto one location


brayniac:
## Summary

`SegmentsBuilder::build` (`crates/segcache/src/segments/builder.rs`) validates a minimum segment size, 8-byte alignment, and that `heap_size` is a non-zero multiple of `segment_size`. Nothing caps `segment_size` at what a `Location`'s offset field can address.

`Location`'s offset field is 20 bits storing `byte_offset >> 3` (`hashtable/location.rs`), i.e. byte offsets up to `(2^20 - 1) * 8` — an 8 MiB ceiling. `pack_location` masks:

```rust
| ((offset >> 3) & location::OFFSET_MASK)
```

with only a `debug_assert!` in front of it. In a release build, a segment larger than 8 MiB wraps the offset field: an item at byte offset `8 MiB + X` packs to the same `Location` as an item at offset `X` in the same segment and incarnation. Two live items then share one location, which every consumer treats as identity — the hashtable's slot compare, `remove`'s expected-location match, `copy_into`'s relink CAS, `get_pinned`'s post-pin revalidation. `check_integrity()` will not see it either; the aliased pair is internally consistent.

## Why file it now

This is pre-existing in kind — the offset field predates the incarnation tag. What makes it worth a ticket is the asymmetry the tagged-locations branch introduced:

- the **id** field now has a hard error. `Segments::from_builder` refuses a heap needing more than `Location::MAX_SEGMENTS` ids, with a `SegmentsError::TooManySegments` that names the limit and tells the operator to increase `segment_size` (`segments/error.rs`).
- the **offset** field still silently masks. And the remedy the id-field error recommends — "increase segment_size" — is precisely the lever that walks a user into the offset field's unenforced ceiling.

The docs meanwhile assert the ceiling as if something enforced it. `Location::MAX_SEGMENTS`' doc computes capacity "2 TiB at the 8 MiB maximum", and the design doc's §2 Layout says the same. There is no 8 MiB maximum in the code; there is an 8 MiB point past which locations alias.

## Suggested fix

[...]

PR #93, opened 2026-08-28
Kani proofs for the bit-packing substrate the concurrency protocols rest on

brayniac:
Item 3 of the concurrency-verification hardening effort (journal entry in-tree; follows #89/#90/#92).

## What

Thirteen `#[kani::proof]` harnesses — exhaustive symbolic proofs over full input domains where the unit tests were hand-picked cases and several load-bearing invariants were English doc-comment arguments:

| Surface | Proven |
|---|---|
| `pack_location`/`unpack_location`/`tag` | roundtrip for every valid (id, generation, 8-aligned offset); injectivity up to the tag projection — with the layout invariant, no two live items share a location word (the #79 silent-wrap class made unreachable); **GHOST unreachability** (previously argued in `MAX_SEGMENTS`' doc comment) |
| `Metadata::pack`/`unpack` | roundtrip + injectivity over every valid (state, links, tag); `State::from_u8` roundtrip |
| `CasToken` / `mix_version` | token roundtrip; version-injectivity for fixed raw — the "odd constant, bijective over u64" comment as a machine-checked fact (what makes CAS tokens observe every in-place increment); the cross-raw non-goal now documented (pigeonhole-impossible for any 64-bit token) |
| TTL `bucket_index` | `< TOTAL_BUCKETS` for **every** i32 — the bound `get_bucket`'s `get_unchecked` rested on via SAFETY comment — plus global monotonicity over positive TTLs |
| keyvalue `numeric_value_pad`/`item_size` | pad < 8, value-slot 8-alignment, size covers-and-aligns, for every klen/olen/vlen |

Two small structure-over-discipline refactors make obligations provable as facts about pure functions rather than claims about `self`: the TTL tier arithmetic moves out of the `&self` method into `bucket_index(secs)`, and the size arithmetic into `item_size_for` (which both value shapes call — so the proof covers the function, not a transcription of its formula).

**Scope honesty:** Kani explores no thread interleavings — loom/shuttle/TSan own that axis — and cannot instantiate the mmap-backed engine. These are deliberate leaf-function proofs; each verifies in well under a second.

## Bite-checks

An offset-shift alias in `pack_location`, removal of `bucket_index`'s tier-4 clamp, and a skewed `Metadata` prev-shift each fail their proof (`VERIFICATION:- FAILED` ×3), restored tree re-verifies 13/13.

## CI

A `kani` job: pinned `KANI_VERSION` behind a cache keyed on it, hand-rolled rather than the upstream action (which pins an old Kani and compiles it from source every run). The install guard is **functional** — it checks the toolchain symlink inside `~/.kani` resolves, because the cache restores `~/.kani` but not `~/.rustup`, so a presence-only guard would leave warm runs with a dangling toolchain (found in review).

## Adversarial review (subagent, static)

Verified safe: every `kani::assume` domain matches the builder-enforced call-site domains exactly (checked against `SegmentsBuilder::build`, `Segments::from_builder`, header field widths); no assume-set is unsatisfiable (each has a concrete witness); the TTL refactor is behavior-identical including the u32→i32 wrap. Fixed from review: the CI warm-cache toolchain hole; the keyvalue Bytes-arm proof was tautological (re-computed the formula inline — now proves the real function via `item_size_for`); three comment-level scope notes (the 64-lifecycle tag window, the cross-raw token non-goal, the monotonicity claim's true strength).

## Validation

13/13 harnesses verified (segcache 11, keyvalue 2, both integrity cfgs); `cargo test --workspace` green; clippy `-D warnings` clean under all-features and default; fmt clean.

https://claude.ai/code/session_01GWv9oJmijF6kkQQh6Ac8h7


brayniac:
CI iteration on the first run: `mix_version_injective_in_version`'s full-width symbolic form (64-bit multiplier circuits) verified in seconds locally but hung CBMC's CNF reduction on the runner for 44+ minutes — and a constructive `(v*K)*K_INV == v` full-width variant stalled the same way locally, so the SAT cost of the multiplier is erratic across environments, full stop. Reworked to what is solver-trivial and still load-bearing: the machine-checked certificate `K & 1 == 1 && K * K_INV == 1 (mod 2^64)` (constant-folded — injectivity follows by ring arithmetic, documented at the harness), plus a 16-bit-bounded symbolic roundtrip as a sanity layer that a wrong K_INV or a broken mix_version expression still fails (bite-checked: a flipped K_INV bit fails the proof). Also exercised in this rerun: the warm-cache functional guard (kani job 45m-timeout → 1m22s).

https://claude.ai/code/session_01GWv9oJmijF6kkQQh6Ac8h7

